PR #1, opened 2020-02-20
Repo syncs

norton120:
This _should_ start syncing master back to BitBucket whenever a PR is merged to master. _should_.


.github/action/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM debian:bullseye-slim
+
+RUN mkdir /app && \
+apt-get update && \
+apt-get install -y git
```
arghmeleg: alpine is nice for this stuff, smaller too:
```
~$ docker run alpine/git git
git: 'git' is not a git command. See 'git --help'.

The most similar command is
	init
~$ docker image ls | grep -E 'alpine|bullseye-slim'
alpine/git                                                                     latest              a518443ef016        16 hours ago        29.8MB
debian                                                                         bullseye-slim       5fd3aad502fc        2 weeks ago         71.9MB
```

.github/action/action.yml
```diff
@@ -0,0 +1,12 @@
+name: bitbucket_sync
+description: 'clones and pushes to bb'
+inputs:
+  bb_api_token:  
+    description: 'auth for user ethanmknox'
```
arghmeleg: `m`

.github/action/Dockerfile
norton120: is the alpine git image solid? it looked a little shifty on DH so I just defaulted to my usual fat whale, but I can give it a shot

arghmeleg: It's solid, more importantly it adheres to your snow theme.

norton120: fact. 🏂 ☃️ 🏔 
I need to figure out how to chain the commands around the run alias so they can all run in the same container, unless I can mount things in the GH workflow between containers? 
since we need to clone the GH master & then push that to bb

Issue #47, opened 2020-07-16
Migrate All of Lookup to SHOW in Snowflake


norton120:
## Why? 
Snowflake object introspection via traditional information_schema lookup is janky, slow, expensive, and occasionally returns the dreaded "too much data" error. Moving all the object lookups to SHOW commands is a better long-term solution.

## Who does this impact? 
- every Snowflake user that wants faster, more reliable SnowShu runs. 

**Note**: will probably require that we clobber several base adapter internals, so this is as good a time as any to standardize and break them out.

eculicny:
Due to the limits on snowflake's SHOW command (max 10K rows) and the ability to select all objects on a server via regex in the configuration, using SHOW COLUMN command at a table/view level could now require numerous extra network calls for large databases. Switching to a schema level could easily run over the 10K row return limit.

I'm currently working on a way to filter the databases and schemas before building the catalog instead of retrieving the details for the entire server/instance before filtering. Querying INFORMATION_SCHEMA by both database and schema should help avoid the "too much data" error.

If there is still too much latency or "too much data", we can add another level down to build the INFORMATION_SCHEMA queries based on the specific tables and views that match the configuration.

PR #50, opened 2021-02-22
Pre filtering DB relations

eculicny:
## Why? 
This generates the performance boost partially expected by #47 

## What Changes?
The base source adapter is now responsible for building the catalog instead of only fetching the relations. This allows filtering to the relation patterns in the config during the fetching of DB info, which reduces the number of calls and amount of metadata handled. Due to the limitations of SHOW, the INFORMATION_SCHEMA is still being queried for table and column level data.

The Catalog class is removed as its functionality is now in the adapter.

The filtering of the relations that is done during the graph build is likely to be unnecessary now, but has not been changed as part of this PR


eculicny:
This PR might also sort of solve issue #6 as well as #4 to an extent (although catalog caching might still be desirable)

snowshu/adapters/source_adapters/base_source_adapter.py
```diff
@@ -25,16 +42,77 @@ def __init__(self, preserve_case:bool=False):
...
+        # get all columns for filtered db/schema
+        catalog = []
+        logger.info('Building filtered catalog...')
+        start_time = time.time()
+        with ThreadPoolExecutor(max_workers=thread_workers) as executor:
+            {executor.submit(accumulate_relations, d, catalog)
+            for d in filtered_schemas}
+
+        logger.info(
+            f'Done building catalog. Found a total of {len(catalog)} relations from the source in {duration(start_time)}.')
```
jedibrillo, line 75: `start_time` will not show the duration just when threads kicked off.

jedibrillo (changes requested):

snowshu/adapters/source_adapters/base_source_adapter.py
eculicny: That `duration` helper function handles both the difference calculation and the formatting. If we want to avoid that, I can create a new one and essentially deprecate the old one.

ankur112358 (approved):

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -252,7 +262,11 @@ def _build_conn_string(self, overrides: Optional[dict] = {}) -> str:
         get_string = "?" + "&".join([arg for arg in get_args])
         return (''.join(conn_parts)) + get_string
 
-    def get_relations_from_database(self, database: str) -> List[Relation]:
+    @overrides
+    def _get_relations_from_database(self, schema_obj: BaseSourceAdapter._DatabaseObject) -> List[Relation]:
+        quoted_database = schema_obj.full_relation.quoted(schema_obj.full_relation.database)  # quoted db name
```
norton120, line 267: this is much much cleaner at the adapter level. We may eventually want to roll this into the base adapter and param the quoting rules once we have several adapters? but this is much more scalable

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -262,29 +276,30 @@ def get_relations_from_database(self, database: str) -> List[Relation]:
...
                                  INNER JOIN
-                                    {database}.INFORMATION_SCHEMA.COLUMNS c
+                                    {quoted_database}.INFORMATION_SCHEMA.COLUMNS c
                                  ON
                                     c.table_schema = m.table_schema
                                  AND
                                     c.table_name = m.table_name
                                  WHERE
-                                    m.table_schema <> 'INFORMATION_SCHEMA'
+                                    m.table_schema = '{case_sensitive_schema}'
```
norton120, line 287: yeah this is way cleaner

snowshu/core/graph.py
```diff
@@ -18,8 +18,8 @@ def __init__(self):
         self.dag: tuple = None
         self.graph: networkx.Graph = None
 
-    def build_graph(self, configs: Configuration,
-                    full_catalog: Tuple[Relation]) -> networkx.DiGraph:
+    # TODO remove extra filtering of relations when building graph
```
norton120, line 21: I'm not sure we ever needed this tbh? seems like an unreachable condition

norton120: nvm I was looking for the broken dag filtering. yeah this can all be vastly simplified now methinks

norton120 (approved):

snowshu/adapters/source_adapters/base_source_adapter.py
norton120: I think it makes sense - `duration(start_time)` assumes you want the duration between start_time and now si?

eculicny: Yeah. It calculates the difference between now and `start_time` within the helper function.

snowshu/adapters/source_adapters/snowflake_adapter.py
eculicny: Yep, this is one of the pieces that should be moved out of the Relation object and into the adapters.

Should I add issues for those so we have record of them (this and the decoupling of pattern matching)?

snowshu/core/graph.py
eculicny: This comment is for the lines like [this one](https://github.com/eculicny/snowshu/blob/pre-filter-relations/snowshu/core/graph.py#L41), but there are more in the `_apply_specifications` function as well.

The broken dag filtering are checks like [this](https://github.com/eculicny/snowshu/blob/pre-filter-relations/snowshu/core/graph.py#L52;L45), unless I'm missing something. If you start with a digraph, it will always be directed. I assume the check should be something like `is_acyclic()` instead since the error message indicates cycles exist.

PR #54, opened 2021-03-03
Better errors messages (input validation)

ankur112358:
This PR improves the error messages that might be encountered while loading config and credentials from replica.yml and credentials.yml. This is done using jsonschema validations.

What Changes?
- [x] validate input
- [x] Tone down logging (snowflake)
- [x] have time.sleep for docker spin up based on constant

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -8,6 +8,7 @@ tabulate==0.8.6
 psycopg2-binary==2.8.4
 scipy==1.4.1
 overrides==3.1.0
+jsonschema==3.2.0
 
 
 #SNOWFLAKE
--- a/snowshu/core/configuration_parser.py
+++ b/snowshu/core/configuration_parser.py
@@ -1,6 +1,9 @@
 from pathlib import Path
 from snowshu.core.utils import fetch_adapter,correct_case
+import os
 import yaml
+import jsonschema
+from jsonschema.exceptions import ValidationError
 from typing import Union, TextIO, List, Optional,Type,Any
 from snowshu.logger import Logger
 from snowshu.configs import DEFAULT_THREAD_COUNT, \
@@ -81,7 +84,8 @@ class Configuration:
 
 class ConfigurationParser:
 
-    def _get_dict_from_anything(self, dict_like_object:Union[str,'StringIO',dict])->dict:
+    def _get_dict_from_anything(self, dict_like_object:Union[str,'StringIO',dict],
+                                **kwargs)->dict:
         """Returns dict from path, io object or dict.  
         
         Returns:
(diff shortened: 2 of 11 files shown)


requirements/base.txt
```diff
@@ -8,6 +8,8 @@ tabulate==0.8.6
 psycopg2-binary==2.8.4
 scipy==1.4.1
 overrides==3.1.0
+ansible==3.0.0
```
norton120, line 11: is there any cheaper way to get the validation than ansible? That's a monster requirement for yaml validation

snowshu/core/configuration_parser.py
```diff
@@ -94,8 +102,49 @@ def _get_dict_from_anything(self, dict_like_object:Union[str,'StringIO',dict])->
...
+        # Load data
+        try:
+            with open(file_path, 'r') as f:
+                loader = AnsibleLoader(f)
+                instance = loader.get_single_data()
+        except FileNotFoundError:
+            loader = None
+            raise
+        except Exception as exc:
+            raise Exception(f'Error when loading YAML config at {file_path} {exc}')
```
norton120, line 129: I'm assuming the loader throws a pretty specific set of exceptions - like KeyErrors or similar right? we should tighten this down a bit if we can.

snowshu/core/configuration_parser.py
```diff
@@ -94,8 +102,49 @@ def _get_dict_from_anything(self, dict_like_object:Union[str,'StringIO',dict])->
...
+        finally:
+            if loader:
+                loader.dispose()
```
norton120, line 132: So.. what does the loader actually _do_ then, if we are just loading the file and then throwing it out? from here it basically looks like it's all wrapped methods from yaml? https://github.com/ansible/ansible/blob/devel/lib/ansible/parsing/yaml/loader.py

snowshu/core/configuration_parser.py
```diff
@@ -94,8 +102,49 @@ def _get_dict_from_anything(self, dict_like_object:Union[str,'StringIO',dict])->
...
+
+        if verify:
```
norton120, line 134: I'm pretty sure we only really need this block

snowshu/templates/replica_schema.yml
```diff
@@ -0,0 +1,115 @@
...
+    "credpath",
+    "long_description",
+    "name",
+    "short_description",
+    "source",
+    "target",
+    "threads",
+    "version"
+  ]
+}
```
norton120, line 115: newline make json mad

norton120 (changes requested):
I am pretty sure we can get away without importing Ansible which would be ideal

snowshu/core/configuration_parser.py
eculicny: I would assume the benefits of using the ansible loader would lie in the templating (jinja) and ability to load their variant of encrypted settings buried in that `AnsibleConstructor` super class.

It's definitely overkill, unless we are planning to mimic their config style, which is probably not worth it at this point.

snowshu/core/configuration_parser.py
eculicny: Slightly related. Is the `verify` flag meant for the case when we have a `yaml` config file that doesn't have a schema file specified for it?

If we are planning on using json schema for everything, I don't think that should ever be the case, and we could have that path error out entirely.

snowshu/core/configuration_parser.py
```diff
@@ -94,8 +98,41 @@ def _get_dict_from_anything(self, dict_like_object:Union[str,'StringIO',dict])->
             try:
                 return yaml.safe_load(dict_like_object.read())
             except AttributeError:
-                with open(dict_like_object,'r') as f:
-                    return yaml.safe_load(f.read())
+                return self._load_and_verify_file(dict_like_object, **kwargs)
```
eculicny, line 101: Is it possible to verify any input here against the jsonschema instead of just the actual file case? Essentially, generate the `instance` object here, and then pass it into the verifying function.

Not sure if that situation is actually possible though.

snowshu/templates/credentials_schema.json
```diff
@@ -0,0 +1,50 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
eculicny, line 2: Do we want a title and description for this schema as well?

eculicny (changes requested):
Looks good! We might be able to simplify the `_get_dict_from_anything` -`_load_and_verify_file` interaction to have the former only handle the loading and the latter only handle the verification.

eculicny (approved):

snowshu/core/configuration_parser.py
ankur112358: Paths can error out entirely if we have a different file names for our input files, I do not think `replica.yml` can be renamed, but the `credpath` in `replica.yml` can contain any name and this can lead to `FileNotFoundError` when looking for schema for the bad credentials file.
https://github.com/Health-Union/snowshu/pull/54/files#diff-ec76ef37dcbb1075645b9ef7cb0021659dad0a01b09ea3908846911999b451f4R111

norton120 (approved):

Issue #43, opened 2020-04-22
SnowShu referential dependency nested queries are repetitive and wasteful. 


norton120:
## Why? 
The current construction of SnowShu queries does not account that the same 10 models in a DAG could have the same 10 identical deps for "include outliers". This creates exponential overhead and grinds the query calculation to crawl. 

## Def of Done: 
Revisit the construction strategy and pair down nested duplicates - I think a final set could be aggregated and unioned one time at the close of the CTE.

PR #56, opened 2021-03-12
Target node in directional relation considered for catalog

ankur112358:
Includes the relations from the second level (target) of the specified relations in replica.yml to create the catalog. 

Ref #43


eculicny (approved):

eculicny (approved):

norton120 (approved):
solid updates

PR #59, opened 2021-03-21
Fix jsonschema validation + Specify schema in replica template

ankur112358:
Issues
- In the `replical.yml`, validation of the first element in a array takes place, but the subsequent elements are ignored
- Integration test `snowflake/test_end_to_end.py::test_finds_n_relations` fails as it picks up additional relations from the schema `SNOWSHU_DEVELOPMENT.TEST_XDB`
- Validation of [realtionships](https://github.com/Health-Union/snowshu/blob/master/snowshu/templates/replica_schema.json#L172-L178) in specified relations uses `anyOf` as a result - if just One relationship is mentioned it works fine but if both are mentioned only one of the correct obj is checked and the other one is not verified.

This PR :-
- [x] fixes the incorrect use of tuple validation and corrects it to list validation
- [x] Restrict the regex in template replica to create limited relations for integration test.
- [x] Fix the incorrect use of `anyOf` to validate both direction of relationships


tests/assets/integration_test_setup.py
```diff
@@ -4,21 +4,46 @@
 import yaml
 import logging
 import os
+import glob
+from pathlib import Path
+from snowshu.adapters.source_adapters.snowflake_adapter import SnowflakeAdapter
+from snowshu.core.models.credentials import Credentials
+
+# raise NotImplementedError('THIS IS A WORK IN PROGRESS, DO NOT USE!')
```
eculicny, line 12: If this is still under construction, we should at least leave the `raise NotImplementedError` in the script.

eculicny (changes requested):

ankur112358:
Tried to split this into sub PRs.
* [Fix jsonschema validation](https://github.com/Health-Union/snowshu/pull/66)
* [Specify schema in replica template](https://github.com/Health-Union/snowshu/pull/67)

PR #57, opened 2021-03-16
Issue 20 pep8 cleanup

ankur112358:
This PR contains small but a number of code cleanups.

- [x] Prospector passes on complete repo
- [x] Create [development docs](https://github.com/Health-Union/snowshu/pull/57/commits/6dd358a9dd4aae8c586ba9a7c964dc7cbf72dcec)
- [x] Update CI to include linting 
- [X] Update CI to run <del>tests after maintainer approval ([done](https://github.com/Health-Union/snowshu/pull/57/commits/143d8a158464c61ec28adee6b3ece791492daad5) using github environment)</del> unit test

PR #60, opened 2021-03-22
Fix json datatype loading

eculicny:
### Problem

JSON datatypes are not being loaded into the replica properly. For snowflake -> postgres, they end up as JSON typed strings in postgres as opposed to actual JSON objects that can be traversed.

When loading snowflake queries via pandas the JSON types get mapped to string. This is intended behavior as noted [here](https://docs.snowflake.com/en/user-guide/python-connector-pandas.html#snowflake-to-pandas-data-mapping).

### Solution

An extra check in the `relation.data.setter` verifies that the JSON types are not `str` typed and converts them if they are. All other types are not modified.

More debug logs around pandas loaded datatypes have also been added in the event there are other unexpected type mappings.


eculicny:
We should merge #57 before merging this PR to avoid constantly playing fix-the-merge-conflict on the other one.

ankur112358 (approved):

Issue #64, opened 2021-03-28
Integration test setup - Standardize table loading process and fix views


ankur112358:
CURRENTLY
* tables for schema `TESTS_DATA` are loaded using the commands in the SQL file `tests/assets/data_types_snowflake_creation.sql` as a result they both have just one row(insert) in them.
* The integration test `tests/integration/snowflake/test_end_to_end.py::test_view` fails.

Def of done
- [ ] passing tests
- [ ] standardize all table loads using csv

PR #67, opened 2021-03-28
Issue 30 - fix dev data setup

ankur112358:
Fixes
- [x] script to setup dev data from csv in snowshu ([one integration test still fails](https://github.com/Health-Union/snowshu/issues/64))
- [x] improves the test `test_finds_n_relations`
    - [x] specify schemas in template schema (used in the test) to not find additional relations
    - [x] improve error messaging of this test


eculicny (commented):
Does this load in the `data_types.csv` to the same table created in the `data_types_snowflake_creation.sql` script?

I'm not sure if it's an issue, but currently the `data_types` table only has the single row from the sql script in it.

eculicny (approved):

PR #68, opened 2021-03-29
Fixed coupling of config options to uuid emulation func

eculicny:
**Problem:**

Building a replica fails when the `pg_extensions` option does not include `uuid-ossp` since the `UUID_STRING()` definition for postgres requires it to be present.

**Solution:**

Add the `CREATE EXTENSION...`  statement to the `UUID_STRING()` SQL file definition so it is always loaded when `UUID_STRING` is loaded.

The references to `uuid-ossp` in the test configurations have been switched to `citext` to avoid collisions during integration tests between the `pg_extensions` option and the emulation functions that require extensions.

### DIFF
--- a/snowshu/adapters/target_adapters/postgres_adapter/functions/UUID_STRING.sql
+++ b/snowshu/adapters/target_adapters/postgres_adapter/functions/UUID_STRING.sql
@@ -1,4 +1,6 @@
 /* emulates snowflake's uuid_string() function for v4 UUIDs, v5 unsupported */
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
 CREATE OR REPLACE FUNCTION UUID_STRING()
 RETURNS TEXT AS
 $$
--- a/snowshu/templates/replica.yml
+++ b/snowshu/templates/replica.yml
@@ -8,7 +8,7 @@ target:
   adapter: 'postgres'
   adapter_args:
     pg_extensions:
-      - uuid-ossp
+      - citext
 source:
   profile: default
   sampling: default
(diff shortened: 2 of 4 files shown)


jedibrillo (approved):

snowshu/adapters/target_adapters/postgres_adapter/functions/UUID_STRING.sql
```diff
@@ -1,4 +1,6 @@
 /* emulates snowflake's uuid_string() function for v4 UUIDs, v5 unsupported */
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
norton120, line 2: Worth maybe at least noting (if not implementing at this time):
bigquery needs `GENERATE_UUID()`
https://cloud.google.com/bigquery/docs/reference/standard-sql/uuid_functions
and then this works for everyone that supports it (there's not a native UUID in RS).

snowshu/templates/replica.yml
```diff
@@ -8,7 +8,7 @@ target:
   adapter: 'postgres'
   adapter_args:
     pg_extensions:
-      - uuid-ossp
+      - citext
```
norton120, line 11: ahh makes sense :+1:

norton120 (approved):
Good stuff, we should note somewhere the one-liner that should get us BQ support as well (same function with an alias)

snowshu/adapters/target_adapters/postgres_adapter/functions/UUID_STRING.sql
eculicny: Currently, snowshu doesn't support bigquery at all as far as I can tell. It looks like you've got bigquery work in branches of your fork, but I don't think any of it was ever merged into upstream.

PR #79, opened 2021-08-24
Fix correct case utility

eculicny:
This PR fixes the changing case of columns with names like `UNIFORM SPACE CASE` and adds tests to verify that. It also addresses the case when the columns start with a number.

**Original Behavior**
When a source DB has the column identifier `UNIFORMLY SPACED COLUMN`, it was mapped to `uniformly spaced column` if the case changed between target and source.

**New Behavior**
When a source DB has the column identifier `UNIFORMLY SPACED COLUMN`, it needs to be preserved since any reference to that column will need to be quoted to handle the spaces.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.5"
 
 services:
   snowshu:
-    image: snowshu
+    image: snowshu-dev
     build:
       context: .
       dockerfile: Dockerfile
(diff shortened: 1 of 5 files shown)


eculicny:
I think we could probably simplify the case problems that snowshu has by ripping out as much case correction as possible and shoring up the quoting behavior. This way, we lean on the SQL dialects to handle casings.

Likely not as simple as that sounds due to the prevalence of case sensitive comparisons and case corrections throughout, but would nicely decouple the case corrections and quoting rules.

jedibrillo (approved):

PR #80, opened 2021-09-17
Support regex in remote relationships

eculicny:
This PR primarily adds support for regex in the upstream/remote relation definitions in the `specified_relations` section of the config. To properly support it, a reworking of the way that the graphs are built was required.

The following config should now be supported:
``` yaml
  - database: my_db
    schema: my_schema
    relation: my_table
    relationships:
      directional:
      - local_attribute: local_attr
        database: ''
        schema: ''
        relation: [upstream]_.*_tables
        remote_attribute: remote_attr
```

**Changes**
* Removed unnecessary refiltering of catalog in `SnowShuGraph`
  * Marked `test_included_and_excluded` as skip since it is no longer relevant due to previous changes where the filtering was happening
* Reworked and refactored `_apply_specifications` to support regex for the upstream relation
  * A side effect of this is a more correct handling of the `''` partitioned wildcard during graph building
* Added errors for no downstream relations, no upstream relations, or defining a many-to-many relationship
* Added tests for those errors and many-to-1 upstream relations

### DIFF
--- a/snowshu/core/graph.py
+++ b/snowshu/core/graph.py
@@ -168,7 +168,7 @@ class SnowShuGraph:
                             relationship[attr] = getattr(downstream_partition[0], attr)
 
                         graph = SnowShuGraph._process_downstream_relation_set(relationship,
-                                                                              downstream_relations,
+                                                                              downstream_partition,
                                                                               graph,
                                                                               available_nodes)
                 # no wildcards present in relationship definition
--- a/snowshu/core/graph.py
+++ b/snowshu/core/graph.py
@@ -4,8 +4,6 @@ import networkx
 
 from snowshu.core.configuration_parser import Configuration
 from snowshu.core.models.relation import (Relation,
-                                          at_least_one_full_pattern_match,
-                                          lookup_single_relation,
                                           single_full_pattern_match)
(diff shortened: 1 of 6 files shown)


snowshu/core/graph.py
```diff
@@ -105,65 +100,142 @@ def _apply_specifications(
...
+                        # populate any wildcard patterns with the appropriate values from first element
+                        for attr in wildcard_attrs:
+                            relationship[attr] = getattr(downstream_partition[0], attr)
+
+                        graph = SnowShuGraph._process_downstream_relation_set(relationship,
+                                                                              downstream_relations,
+                                                                              graph,
+                                                                              available_nodes)
+                # no wildcards present in relationship definition
+                else:
```
jedibrillo, line 175: could you remove the `else` and only have one call to `SnowShuGraph._process_downstream_relation_set()`

eculicny: No, but that's because I screwed up line 170. The other call needs to use `downstream_partition` not `downstream_relations` to handle the wildcard behavior properly

eculicny: Fixed with test.

jedibrillo (approved):

PR #86, opened 2021-10-28
Implemented a docker build step for the target image (to avoid finalization steps)

fleischer89:


snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -193,7 +192,9 @@ def _get_data_type(self, source_type: str) -> DataType:
     @staticmethod
     def _build_snowshu_envars(snowshu_envars: list) -> list:
         """helper method to populate envars with `snowshu`"""
-        return [f"{envar}=snowshu" for envar in snowshu_envars]
+        envars = [f"{envar}=snowshu" for envar in snowshu_envars]
+        envars.append(f"PGDATA=/{DOCKER_REMOUNT_DIRECTORY}")
+        return envars
```
eculicny, line 197: Let's refactor the envar setup to work the way we were discussing earlier, where the keys and values are set by the more concrete type of the adapter.

Only the postgres adapters are going to care about the `PGDATA` envar.

eculicny (changes requested):
Since we need to use the envars to get this working, lets refactor those a little bit to move the values out of the `base_target_adapter`

Aesthet (approved):
Looks good to me

eculicny (approved):

PR #87, opened 2021-11-02
Added graph_difference method to SnowShuGraph class to handle increme…

Aesthet:
…ntal replica builds. Added needed flag for incremental builds into the main module. Updated with eq/hash functions classes like Relation and Materialization. Updated _safe_query method of BaseSQLAdapter to accept database as an optional parameter to use as connection's database_override. Updated mock imports, because mock is now part of the Python standard library, available as unittest.mock in Python 3.3 onwards.


snowshu/core/main.py
```diff
@@ -82,16 +82,23 @@ def source(filename):
               default=None,
               help="Overrides the replica name found in replica.yml")
 @click.option(
-    '--barf',
+    '--barf', '-b',
     is_flag=True,
     help="outputs the source query sql to a local folder snowshu_barf_output")
+@click.option(
```
eculicny, line 88: Probably should save the parameters implementation for the subsequent ticket as discussed

snowshu/core/graph.py
```diff
@@ -1,28 +1,89 @@
...
+
+        """
+        if isinstance(source_catalog, (tuple, list, set)):
+            source = list(source_catalog)
+        elif isinstance(source_catalog, SnowShuGraph):
+            source = list(source_catalog.graph.nodes)
+        else:
+            source = list(source_catalog.nodes)
+        non_isolated_relations = []
+        for relation in configs.specified_relations:  # noqa pylint: disable=too-many-nested-blocks
```
eculicny, line 50: At the expense of comparing raw catalogs, we should use the graph's edges to determine the connections between nodes.

eculicny (dismissed):
As discussed, we can save the command options for follow up work, and simplify the connected nodes check with the existing graph.

If we ever need to compare the raw source and target catalogs while taking expected connections into account, we should probably refactor the logic in the `build_graph` function.

Dockerfile
```diff
@@ -1,6 +1,6 @@
 ## docker image is intended for development environments only.
 ## to use snowshu in production install the package via pip
-FROM python:3.7-slim-buster
+FROM python:3.8-slim-buster
```
Aesthet, line 3: This is not needed as I removed usage of walrus operators - should I revert this change?

eculicny: Let's revert it for now since the `setup.py` seems to only expect python version 3.7. We can work on upgrading/updating that dependency later.

eculicny (commented):
Looks good! Need to wait until we can verify that the integration tests are passing before approving.

Dockerfile
Aesthet: done

jedibrillo (approved):

fleischer89 (approved):

PR #88, opened 2021-11-03
Add an override to the base image used for creating a snowshu replica

fleischer89:


snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -125,16 +126,20 @@ def load_data_into_relation(self, relation: Relation) -> None:
             raise exc
         logger.info('Data loaded into relation %s', relation.quoted_dot_notation)
 
-    def initialize_replica(self, source_adapter_name: str) -> None:
+    def initialize_replica(self, 
+                           source_adapter_name: str,
+                           target: str) -> None:
```
eculicny, line 131: Since we use "source" and "target" in a bunch of different circumstances, we should make sure this arg (and probably most other references) indicates that it's the target docker image as opposed to some other target type.

snowshu/core/main.py
```diff
@@ -102,15 +106,18 @@ def create(replica_file: click.Path,
         exists=True),
     default=REPLICA_DEFAULT,
     help="where snowshu will look for your replica configuration file, default is ./replica.yml")
+@click.option('--target',
```
eculicny, line 109: There's no need to include an override option to the CLI. We should only use the override in the case of an incremental build, which will be taken care of by a future `--incremental <target_image>` option

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -164,8 +170,10 @@ def finalize_replica(self) -> str:
         logger.info('Finalized replica image %s', self.replica_meta["name"])
         return replica_image.tags[0]
 
-    def _generate_credentials(self) -> Credentials:
-        host = DOCKER_TARGET_CONTAINER if IS_IN_DOCKER else 'localhost'
+    def regenerate_credentials(self, host) -> Credentials:
```
eculicny, line 173: Why do we need to regenerate the credentials? Is there something that can't be moved to the `__init__` function?

eculicny (changes requested):
It looks like most of the changes are for the `DOCKER_TARGET_CONTAINER`, which shouldn't need to be changed since that probably only affects the name of the target running container, and we don't want to assume the user has started the override image as a container already.

Changing the base image should require modifying how the target adapters deal with the `DOCKER_IMAGE` value located [here](https://github.com/Health-Union/snowshu/blob/master/snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py#L19).

snowshu/adapters/target_adapters/base_target_adapter.py
fleischer89: The `regenerate_credentials()` method is used to regenerate if a target image is passed  which might be different from the default target image

snowshu/core/docker.py
```diff
@@ -87,28 +87,29 @@ def startup(self,   # noqa pylint: disable=too-many-arguments
...
+                base_image: str,
                 protocol: str = "tcp") -> docker.models.containers.Container:   # noqa pylint: disable=unused-argument
 
         container = self.get_stopped_container(
             image,
             start_command,
             envars,
             port,
-            name=DOCKER_TARGET_CONTAINER,
+            name=base_image,
```
eculicny, line 98: We should probably leave the container name alone for consistency and a more convenient short hand than using the full image names for the container name.

snowshu/core/main.py
```diff
@@ -81,18 +81,22 @@ def source(filename):
 @click.option('--name',
               default=None,
               help="Overrides the replica name found in replica.yml")
+@click.option('--target',
```
eculicny, line 84: As mentioned before, we don't need this argument for the time being and it should be removed.

snowshu/core/replica/replica_factory.py
```diff
@@ -43,8 +46,11 @@ def _execute(self,
             return "No relations found per provided replica configuration, exiting."
 
         # TODO replica container should not be started for analyze commands
+        target = target if target is not None else DOCKER_TARGET_CONTAINER
```
eculicny, line 49: Since `initialize_replica` expects the image name and not the container name, the default `target` shouldn't have a value since the base image is determined by the concrete target adapter.

eculicny (changes requested):
We should also add an integration test case to verify that the using a different image works. The output image of the `test/integration/snowflake/test_end_to_end.py` tests should be able to be used for that test case.

fleischer89:
This PR has changes from PR [(Feature/data 1904 #89)](https://github.com/Health-Union/snowshu/pull/89) merged into it

snowshu/core/main.py
```diff
@@ -105,11 +110,19 @@ def create(replica_file: click.Path,
 @click.option('--barf', '-b',
               is_flag=True,
               help="outputs the source query sql to a local folder snowshu_barf_output")
-def analyze(replica_file: click.Path, barf: bool):
+@click.option(
+    '--incremental', '-i',
+    is_flag=True,
+    help="creates relations and loads data only for new entries found in replica.yml, "
+         "which are not already present in target replica image")
```
eculicny, line 117: Unless there is a specific reason to include this flag for the analyze command, we should probably exclude it, and leave `replica.incremental` as `None` or `False`.

Alternatively, if we want to keep behavior parity, this option should take an image name so that the graph difference can be calculated and only those relations end up being analyzed.

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -189,6 +197,13 @@ def _get_data_type(self, source_type: str) -> DataType:
                 '%s adapter does not support data type %s.', self.CLASSNAME, source_type)
             raise err
 
+    @staticmethod
+    def _build_snowshu_envars(snowshu_envars: list) -> list:
+        """helper method to populate envars with `snowshu`"""
+        envars = [f"{envar}=snowshu" for envar in snowshu_envars]
+        envars.append(f"PGDATA=/{DOCKER_REMOUNT_DIRECTORY}")
```
eculicny, line 204: With `PGDATA` in the postgres adapter, it can be removed here.

eculicny (changes requested):
Looks good! Only some minor changes.

Does this PR already include everything from #89? So that once it's merged we can just close the other one?

fleischer89:
> Looks good! Only some minor changes.
> 
> Does this PR already include everything from #89? So that once it's merged we can just close the other one?

Yes it does

snowshu/core/main.py
fleischer89: > Unless there is a specific reason to include this flag for the analyze command, we should probably exclude it, and leave `replica.incremental` as `None` or `False`.
> 
> Alternatively, if we want to keep behavior parity, this option should take an image name so that the graph difference can be calculated and only those relations end up being analyzed.

We are taking it out for now. We can spin off the graph difference calculation in new ticket

Aesthet: Yeah - I was thinking whether we should leave this here. Asked Paul to remove it

snowshu/adapters/base_sql_adapter.py
```diff
@@ -154,7 +154,7 @@ def _build_conn_string_partial(
             {USER, PASSWORD, HOST, DATABASE, }
         )
 
-    def build_catalog(self, patterns: Iterable[dict], thread_workers: int = 1) -> Tuple[Relation]:
+    def build_catalog(self, patterns: Iterable[dict], thread_workers: int) -> Set[Relation]:
```
Aesthet, line 156: I deleted default as agreed with @eculicny

snowshu/core/printable_result.py
```diff
@@ -36,8 +36,8 @@ def graph_to_result_list(graphs: nx.Graph) -> list:
             for relation in graph.nodes:
                 deps = len(nx.ancestors(graph, relation))
                 deps = " " if deps == 0 else str(deps)
-                target_sample_size = relation.population_size if\
-                    (relation.unsampled or relation.population_size < relation.sampling.size)\
+                target_sample_size = relation.population_size if \
+                    relation.unsampled or (relation.population_size < relation.sampling.size) \
```
Aesthet, line 40: @eculicny I Guess this parentheses placement makes more sense?

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -15,7 +15,7 @@
 
 class PostgresAdapter(BaseTargetAdapter):
     name = 'postgres'
-    dialect = 'postgres'
+    dialect = 'postgresql'
```
Aesthet, line 18: This fixes warning:
```
tests/integration/snowflake/test_analyze.py::test_analyze_unsampled
  /usr/local/lib/python3.7/site-packages/sqlalchemy/util/langhelpers.py:263: SADeprecationWarning: The 'postgres' dialect name has been renamed to 'postgresql'
    loader = self.auto_fn(name)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

snowshu/core/printable_result.py
eculicny: I also like the updated version better since it more indicative of the operands for the operators. If it makes sense to everyone else keep it in.

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
eculicny: Yeah, probably worth pulling the trigger on this change. Originally I was always waiting for the point that the pins would be updated to make this change, but that day never seemed to arrive.

eculicny (approved):

PR #89, opened 2021-11-10
Feature/data 1904

Aesthet:
Added incremental image argument and needed logic atop graph difference. Merged changes made in scope of 'Add an override to the base image used for creating a snowshu replica' made by Paul


Aesthet:
New PR will be created from Paul's fork with integrated logic from both current PRs.

PR #90, opened 2021-11-12
Reworked get_graphs to use networkx methods instead of custom logic, …

Aesthet:
…renamed it to get_connected_subgraphs.


snowshu/core/graph.py
```diff
@@ -298,49 +298,28 @@ def _process_downstream_relation_set(
...
-
-        dags = [networkx.DiGraph() for _ in range(len(isodags))]
-        for graph, node in zip(dags, isodags):
-            graph.add_node(node)
-
-        for collection in node_collections:
-            dags.append(networkx.subgraph(self.graph, collection))
+            raise ValueError('Graph must be built before SnowShuGraph can get graphs from it.')
+
+        graph = self.graph.to_directed()
```
eculicny, line 311: Is there a reason we need to specify `to_directed` here? The graph that was built should already be a digraph (that disallows multiple edges between two nodes)

eculicny (commented):
Need to resolve some merge conflicts, but looks good!

snowshu/core/graph.py
eculicny: That makes sense. We should probably fix those tests and remove the graph modification in the function. At no point should we be creating an undirected graph currently.

The tests were likely created using an undirected graph since the original `get_graphs` function called `graph.to_undirected()`

eculicny (approved):

PR #91, opened 2021-11-19
Added polymorphic relationships to documentation of replica.yml

fleischer89:


docs/source/user_documentation/replica_dot_yaml_file.rst
```diff
@@ -288,6 +389,55 @@ The *specified relation* is ``SNOWSHU_DEVELOPMENT.SOURCE_SYSTEM.ORDER_ITEMS``. W
...
+              - database: SNOWSHU_DEVELOPMENT
+                schema: POLYMORPHIC_DATA
+                relation: CHILD_TYPE_2_ITEMS
+                override_value: type_2
+
+
+The *specified relation* is ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.PARENT_TABLE`` which relates with a child relations ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.(?i)^CHILD_TYPE_[0-9]_ITEMS$``. When SnowShu builds this replica:
+
+- All the records in ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.PARENT_TABLE`` will be records with a ``CHILD_ID`` of ``local_type_attribute``, ``CHILD_TYPE``, found in the ``ID`` of any tables that match ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.(?i)^CHILD_TYPE_[0-9]_ITEMS$`` (example: ``ID`` in ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.CHILD_TYPE_3_ITEMS``).
+- Records found in ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.CHILD_TYPE_2_ITEMS`` will be records with a ``ID`` of type ``type_2`` (using the ``override_value`` of ``local_type_overrides``) found in the ``CHILD_ID`` of ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.PARENT_TABLE``.
```
eculicny, line 438: Polymorphic relationships are only enforced when sampling the parent table. So we have no guarantees that the child table records will be referenced by the parent table records.

docs/source/user_documentation/replica_dot_yaml_file.rst
```diff
@@ -288,6 +389,55 @@ The *specified relation* is ``SNOWSHU_DEVELOPMENT.SOURCE_SYSTEM.ORDER_ITEMS``. W
...
+            local_type_overrides:
+              - database: SNOWSHU_DEVELOPMENT
+                schema: POLYMORPHIC_DATA
+                relation: CHILD_TYPE_2_ITEMS
+                override_value: type_2
+
+
+The *specified relation* is ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.PARENT_TABLE`` which relates with a child relations ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.(?i)^CHILD_TYPE_[0-9]_ITEMS$``. When SnowShu builds this replica:
+
+- All the records in ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.PARENT_TABLE`` will be records with a ``CHILD_ID`` of ``local_type_attribute``, ``CHILD_TYPE``, found in the ``ID`` of any tables that match ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.(?i)^CHILD_TYPE_[0-9]_ITEMS$`` (example: ``ID`` in ``SNOWSHU_DEVELOPMENT.POLYMORPHIC_DATA.CHILD_TYPE_3_ITEMS``).
```
eculicny, line 437: I think this is right, but could probably be clearer. Here's one stab at it that should be correct, but probably could also be phrased better

"All records in the parent table will be records that have a `CHILD_TYPE` (the `local_type_attribute` column) value that matches the child table name (or the `local_type_override` for the child table if used), and a `CHILD_ID` value that matches `ID` value in that matching table."

eculicny (changes requested):

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -135,9 +137,21 @@ def initialize_replica(self, source_adapter_name: str, override_image: str = Non
             override_image: the name of incremental image to initialize,
                 if specified will override default image
         """
-        if override_image:
-            self.__class__.DOCKER_IMAGE = override_image
         try:
+            if override_image:
+                shdocker = SnowShuDocker()
+                images = shdocker.client.images.list(name=override_image)
+                if len(images) > 0:
```
eculicny, line 144: Since we've got the generic `except` block, we can probably skip this check and log the found images before proceeding

If we leave this `if` statement in, then `else`  block would need to properly raise an error

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -135,9 +137,21 @@ def initialize_replica(self, source_adapter_name: str, override_image: str = Non
...
+                    image_commands = []
+                    for item in images[0].history():
+                        if ("postgres" in item["CreatedBy"]) or ("PGDATA" in item["CreatedBy"]):
+                            image_commands.append(item["CreatedBy"])
+                    if len(image_commands) > 0:
+                        self.__class__.DOCKER_IMAGE = override_image
+                    else:
+                        logger.error(f"The override image is not a Postgres image: {override_image}")
```
eculicny, line 152: This `else` block also should raise an error. If there are problems with the image passed to `--incremental`, snowshu should probably abort instead of falling back to a full replica build.

eculicny (changes requested):

eculicny:
Do the integrations tests pass for this? They don't seem to work locally for me since the new test, `test_incremental_build_with_override_image` fails when it can't find `/app/replica.yml`.

fleischer89:
> Do the integrations tests pass for this? They don't seem to work locally for me since the new test, `test_incremental_build_with_override_image` fails when it can't find `/app/replica.yml`.

I've updated the test to pass the `replica.yml` file so we don't encounter this

PR #93, opened 2021-11-30
Implemented verification test for incremental build with override image

fleischer89:


eculicny (commented):
Mostly looks good except for the test.

Are we expecting the running command to succeed? Currently, the given image to the `--incremental` flag doesn't exist. It should be 'snowshu_replica_integration-test:latest`. 

If we're expecting it to fail, we shouldn't raise an error on line 411 on a failure exit_code.

In any case, the string on 413 doesn't seem to appear any where in the code. The closest I've found is [here](https://github.com/Health-Union/snowshu/blob/master/snowshu/adapters/target_adapters/base_target_adapter.py#L143)

eculicny (approved):

PR #94, opened 2021-11-30
Bug Fix: FDWs not showing with incremental build

fleischer89:


snowshu/core/replica/replica_factory.py
```diff
@@ -39,6 +40,8 @@ def _execute(self,
             self.config.name = name
 
         graph.build_graph(self.config)
+        cross_db_relations = list(graph.graph.nodes) if graph.graph is not None and \
+            not isinstance(graph.graph, Mock) else None
```
eculicny, line 44: The application code shouldn't need to branch for testing. As `Mock` isn't used internally to snowshu, it shouldn't be handled here.

Also, if this method fixes the problem, we shouldn't have any cross db references outside of the `if self.config.source_profile.adapter.SUPPORTS_CROSS_DATABASE` block. This variable should be renamed.

eculicny (changes requested):
Unless I'm missing something, I don't see how this actually fixes the foreign data wrapper not working. 

This generates the full list of relations in graph _before_ any are removed when running an incremental build, and that set is used to create the cross database relations.

But the problem was that the newly added relations during an incremental build didn't have FDW versions added during the `enable_cross_database` function. Those newly added relations should already have been included in the `relations` set that is included on line 74 (replica_factory).

Can we add a failing test case to ensure that this is fixed

fleischer89:
> Unless I'm missing something, I don't see how this actually fixes the foreign data wrapper not working.
> 
> This generates the full list of relations in graph _before_ any are removed when running an incremental build, and that set is used to create the cross database relations.
> 
> But the problem was that the newly added relations during an incremental build didn't have FDW versions added during the `enable_cross_database` function. Those newly added relations should already have been included in the `relations` set that is included on line 74 (replica_factory).
> 
> Can we add a failing test case to ensure that this is fixed

The problem was that the database (`data_lake` in the case of the example replica) was not being passed to `enable_cross_database` for the importing of the **FDW**s so it was not able to import them that's why I went for this implementation to resolve this.

eculicny:
> The problem was that the database (`data_lake` in the case of the example replica) was not being passed to `enable_cross_database` for the importing of the **FDW**s so it was not able to import them that's why I went for this implementation to resolve this.

I was missing something, sorry about that. We should just have the `enable_cross_database` inspect the running instance to get the dbs/schemata (non-default ones) to setup FDWs for instead of expecting them to be passed as discussed.

tests/integration/snowflake/test_end_to_end.py
```diff
@@ -400,3 +393,13 @@ def test_using_different_image(end_to_end):
     target_container.reload()
     assert target_container.status == 'running'
     target_container.remove(force=True)
+
+
+def test_incremental_build_with_override_image(end_to_end):
```
eculicny, line 398: Except for this failing test, the rest of the PR looks fine.

eculicny (approved):

PR #95, opened 2021-12-06
Implemented creation of postgres extensions (post-processing step)

fleischer89:


eculicny (changes requested):
Need to fix merge conflicts (i.e. remove that bad test).

Also, we should remove the creation of the extensions during data loading here: https://github.com/Health-Union/snowshu/blob/master/snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py#L102

eculicny (approved):

PR #96, opened 2021-12-14
Added thread handler to graph_set_runner: retry mechanism + fail prog…

Aesthet:
…ram if at least one thread can't be finished successfully.

### DIFF
--- a/snowshu/core/graph_set_runner.py
+++ b/snowshu/core/graph_set_runner.py
@@ -4,7 +4,7 @@ import shutil
 import time
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
-from typing import List
+from typing import List, Tuple
 
 import networkx as nx
 
@@ -35,7 +35,7 @@ class GraphSetRunner:
         self.barf = None
 
     def execute_graph_set(self,     # noqa pylint: disable=too-many-arguments
-                          graph_set: List[nx.Graph],
+                          graph_set: Tuple[nx.Graph],
                           source_adapter: BaseSourceAdapter,
                           target_adapter: BaseTargetAdapter,
                           threads: int,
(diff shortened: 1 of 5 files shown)


snowshu/core/graph_set_runner.py
```diff
@@ -70,10 +70,59 @@ def make_executables(graphs) -> List[GraphExecutable]:
...
+                                # if exception in thread happened - record reference to it and append
+                                # actual executable to a list for re-execution
+                                error = future.result
+                                logger.warning("Concurrent thread finished work with exception:\n%s: %s",
+                                               str(exception.__class__),
+                                               str(exception))
+                                re_executables.append(executable)
+                        else:
+                            wait_list.append((future, executable))
+                    while wait_list:
```
Aesthet, line 96: this can be reworked not to have duplicate code, but will require additional nested function and will cause complexity raise

eculicny: Would it make sense to only have a nested function for this duplicated code instead of the `process_executables` and `make_executables` functions, and extract `process_executables` into its own separate function (with adjusted inputs)?

That way, we'd only have one level of nesting inside `process_executables` and no nesting in `execute_graph_set`?

eculicny (approved):

snowshu/core/graph_set_runner.py
```diff
@@ -59,21 +60,53 @@ def execute_graph_set(self,     # noqa pylint: disable=too-many-arguments
...
+                for future, executable in zip(futures.copy(), executables.copy()):
+                    if future.done():
+                        futures.remove(future)
+                        executables.remove(executable)
+                        if exception := future.exception():
+                            error = future.result
+                            logger.warning("Concurrent thread finished work with exception:\n%s: %s",
+                                           str(exception.__class__),
+                                           str(exception))
+                            re_executables.append(executable)
```
Aesthet, line 99: looks like I am appending executables to be re-executed here, and also have an assignment on line 108, but I never delete / recreate it...
will update

eculicny: I think a `clear()` call and a `copy()` on line 108 would be needed to clean it up properly.

The assignment on line 108 should have the `executables` and `re_executables` variables point to the same list from that point forward which gets modified during `execute_with_retry` function to only contain the objects that need to be retried. So even though the `re_executables` list doesn't get cleared, it gets cleaned up on successes by line 93's `remove()` calls.

If we only add the `clear()` call on line 85, I don't think anything will actually get retried, and they will have all silently failed since the `executables` list will be empty after the `clear()` call.

Aesthet: yeah, tested it - adding clear and copy is enough

PR #98, opened 2022-01-28
Added path separator to the docker.sock path

ArtemHU:


.gitignore
```diff
@@ -11,6 +11,8 @@ build/
 *.log.*
 snowshu_barf_output/
 keyfile.json
+credentials.yml
+replica.yml
```
eculicny, line 15: We do still have two files named `replica.yml` and `credentials.yml` in the `templates` directory, so we should probably not ignore them for the time being.

eculicny (commented):
Looks good other than rolling back the small change to `.gitignore`

eculicny (approved):

.gitignore
ArtemHU: Update the file

PR #99, opened 2022-01-31
DATA-2025: Fixed PyPI deployment pipeline

fleischer89:


eculicny:
We should probably move the publish action to its own action yml file so that it isn't also executed on PRs and branch pushes.

Likely should have it only run when triggered manually since we don't want to create a new release for every single minor fix that happens. Alternatively, if there is an easy way to trigger a pipeline on a new tag being pushed/updated, we could use that too.

.github/workflows/live_deployment.yml
```diff
@@ -0,0 +1,25 @@
...
+
+  live-deployment-to-pypi:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Setup template credentials
+      run: cp snowshu/templates/credentials.yml tests/assets/integration/credentials.yml
```
eculicny, line 16: I don't think it would get picked up, but we don't need to add the test credentials for publishing pipelines.

.github/scripts/deploy_to_pypi.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/sh
 
+action=$1
 # Upload package to PyPI using github secrets
-twine upload -u __token__ -p "$TWINE_PASSWORD" dist/*
\ No newline at end of file
+if [[ $action == "live_deploy" ]]; then
+    twine upload -u __token__ -p "$PYPI_KEY" dist/*
+else
+    twine upload -u __token__ -p "$TEST_PYPI_KEY" --repository testpypi dist/*
```
eculicny, line 8: We can probably consolidate `PYPI_KEY` and `TEST_PYPI_KEY` into a single env var that is passed in the workflows. This way, we don't pass a valid production key for pypi when running test deploys.

requirements/base.txt
```diff
@@ -10,6 +10,7 @@ scipy==1.7.3
 overrides==3.1.0
 jsonschema==3.2.0
 tenacity==7.0.0
+twine==3.7.1
```
eculicny, line 13: Since we're using the dev docker image, we can add this requirement to `dev.txt` instead of `base.txt`

eculicny (changes requested):

eculicny (changes requested):
Can we consolidate the environment variables [here](https://github.com/Health-Union/snowshu/pull/99/files#r802013337) so we aren't passing around live credentials to test workflows.

eculicny (approved):

Aesthet (approved):
looks good to me

PR #100, opened 2022-02-02
DATA-2026: Added publishable docker image to repository

fleischer89:


 docker_image/snowshu.dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.7
```
eculicny, line 1: We should bump this up to 3.8 since we've added the walrus operator into the code which was first available in 3.8 I believe.

 docker_image/snowshu.dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+RUN pip3 install snowshu>=0.0.3 && \
```
eculicny, line 3: Not sure if there is a more convenient way of specifying the version of snowshu to install, but as long as the image is built after the push to pypi, this should pull the latest version in.

Might be worth a quick investigation into what types of parameterization github actions supports for docker builds.

 docker_image/README.md
```diff
@@ -0,0 +1,18 @@
+
+# Docker SnowShu
+
+A simple image for running SnowShu in a container. 
+
+*Note:* SnowShu uses docker, so when you run the container make sure to add it to the snowshu network with
```
eculicny, line 6: Since snowshu manages another container, the description should probably include a note about making sure the docker socket is included as a volume as well (even though it's included in the example run command)

eculicny (changes requested):

 docker_image/snowshu.dockerfile
Aesthet: true

 docker_image/snowshu.dockerfile
Aesthet: Since the latest version is what we want we just need to perform `pip3 install snowshu`
https://docs.python.org/3/installing/index.html#basic-usage

eculicny (changes requested):
Can we add a note about the docker socket file as mentioned [here](https://github.com/Health-Union/snowshu/pull/100/files#r797976148)

eculicny (approved):

ArtemHU (approved):

PR #103, opened 2022-02-17
 DATA-2306: Refactored quoting and casing across relations, adapters and associated tests

fleischer89:


snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -41,7 +41,7 @@ class SnowflakeAdapter(BaseSourceAdapter):
     REQUIRED_CREDENTIALS = (USER, PASSWORD, ACCOUNT, DATABASE,)
```
ArtemHU, line 41: It seems to me that we need to update the description of the class

snowshu/core/models/relation.py
```diff
@@ -111,8 +111,8 @@ def relation(self, value: str) -> None:
     @staticmethod
     def quoted(val: str) -> str:
         """Returns quoted value if appropriate."""
-        return f'"{val}"' if all({val.isupper(),
-                                  val.islower(), }) and ' ' not in val else val
+        return f'"{val}"' if any({val.isupper(),
```
ArtemHU, line 114: looks like an issue in the original code

ArtemHU (approved):

snowshu/core/models/relation.py
eculicny: Yeah, that first condition would never be able to evaluate as True.

However, we should remove this function entirely since we're moving the quoting functionality into the adapters.

eculicny (changes requested):
Does this actually keep the default case of the tables in the postgres database? The integration tests in `snowflake/test_end_to_end.py` are all failing with non existing DBs which seem to be all upper case.

I didn't think the target adapter changes the cases of the values it was given, and I think this was the reason for the Snowflake adapter being lower cased. The table from snowflake `"TABLE1"` gets loaded into `"TABLE1"` into postgres, but the intended behavior is that the objects in postgres have the default casing there if they have the default casing in snowflake which would be `"table1"`.

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -198,6 +201,10 @@ def _get_data_type(self, source_type: str) -> DataType:
                 '%s adapter does not support data type %s.', self.CLASSNAME, source_type)
             raise err
 
+    def quoted_dot_notation(self, rel: Relation) -> str:
```
eculicny, line 204: Since the `quoted_dot_notation` function is the same everywhere, we can probably put this back onto the `BaseSQLAdapater` base class instead of each individual class.

snowshu/core/models/relation.py
```diff
@@ -108,25 +103,11 @@ def relation(self) -> str:
...
-        """Returns quoted value if appropriate."""
-        return f'"{val}"' if any({val.isupper(),
-                                  val.islower(), }) and ' ' in val else val
-
     def scoped_cte(self, string: Optional[str] = None) -> str:
         """ returns a CTE name scoped to the relation.
             If _string_ is provided, this will be suffixed to the name."""
         return "__".join([self.database, self.schema, self.name, string])
 
-    def typed_columns(self, data_type_mappings: dict) -> str:
```
eculicny, line 122: I don't recognize this function, but I'm assuming this function isn't used anywhere since we're removing it.

snowshu/adapters/base_sql_adapter.py
```diff
@@ -218,8 +218,7 @@ def _get_filtered_schemas(self, filters: Iterable[dict]) -> List[_DatabaseObject
         # get all schemas in all databases
         filtered_schemas = []
         for db_rel in filtered_databases:
-            schemas = self._get_all_schemas(
-                database=db_rel.quoted(db_rel.database))
+            schemas = self._get_all_schemas(database=db_rel.database)
```
eculicny, line 221: Do we need to preserve the `quoted` call here?

eculicny (commented):
Integrations tests are still failing. Taking a look to see if I can find the root cause of the failure to build a replica in the tests

Added a few comments on general clean-up/consolidation of some of the code.

eculicny:
There are also some references to `relation.quoted_dot_notation()` in the base target adapter that were missed as well [here](https://github.com/Health-Union/snowshu/pull/103/files#diff-16bf5f4c4e5fbc190573fd6e453535fde679c15b3cc3214788d9b91390084360L82)

snowshu/core/models/relation.py
fleischer89: Yes. It's not being used anywhere

snowshu/adapters/base_sql_adapter.py
fleischer89: The `quoted` call is applied on the database in the overridden method in `SnowflakeAdapter`

eculicny: Got it. This change might be related to the issues with the PostgresAdapter on the target side since they both inherit from the `BaseSQLAdapter` class.

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -113,6 +115,10 @@ def create_database_if_not_exists(self, database: str) -> str:
         return database
 
     def create_schema_if_not_exists(self, database: str, schema: str) -> None:
+        database = database if self.preserve_case \
+            else correct_case(database, self.DEFAULT_CASE == 'upper')
+        schema = schema if self.preserve_case \
+            else correct_case(schema, self.DEFAULT_CASE == 'upper')
```
eculicny, line 126: There's already a function to encapsulation this behavior in `BaseSQLAdapter`, `_correct_case`. If we need to check the `preserve_case` flag or make any sort of case correction, we can just use that function.

eculicny:
The integration tests still fail locally for me, and I think it's due to some references to `rel.database` and `rel.schema` that aren't case corrected on the target side of things (probably in the `BaseTargetAdapter` or even in `BaseSQLAdapter`).

Something that may make thing simpler would be to make sure `quoted_dot_notation` returns the case corrected DB object name. So we could move all of that to the `BaseSQLAdapter` class like below.

Then each individual adapter that inherits from the base will already have `quoted_dot_notation` defined, and only need to implement their flavor of quoting in the `quoted` function which requires an override.

```
    @staticmethod
    def quoted(val: str) -> str:
        raise NotImplementedError()

    def quoted_dot_notation(self, rel: Relation) -> str:
        case_corrected_database = self._correct_case(rel.database)
        case_corrected_schema = self._correct_case(rel.schema)
        case_corrected_name = self._correct_case(rel.name)
        return '.'.join([self.quoted(val) for val in
            (case_corrected_database, case_corrected_schema, case_corrected_name)])
```

snowshu/core/graph.py
```diff
@@ -282,7 +283,8 @@ def _process_downstream_relation_set(
                 f'as they are usually unintended side effects of lenient regex.'
             )
         # check to make sure found upstream relation is not a view
-        view_relations = [r.quoted_dot_notation for r in upstream_relations if r.is_view]
+        adapter = SnowflakeAdapter()
+        view_relations = [adapter.quoted_dot_notation(r) for r in upstream_relations if r.is_view]
```
eculicny, line 287: I just noticed this change while looking at the `preserve_case` PR. The core snowshu code shouldn't need to have any idea what the adapter's implementation is.

I don't know why we are looking for the quoted notation here as we are only concerned with the existence of views, not the exact names of them. Probably nice print output.

We can eliminate this use of `quoted_dot_notation` here and the references to the snowflake adapter.

eculicny (approved):
I think we're good now.

PR #105, opened 2022-02-28
DATA-2305: Implemented `preserve_case` flag to handle casing

fleischer89:


tests/conftest_modules/test_configuration.py
```diff
@@ -65,6 +65,11 @@
             ]
         },
         "include_outliers": True,
+        "preserve_case": {
+            "database": True,
+            "schema": True,
+            "relation": True
+        },
```
eculicny, line 72: For the existing tests, we should leave this as the default or explicitly set to all `False`.

Additional tests for the preserving case functionality can pick up this config and modify the `"preserve_case"` flag to verify the added behavior.

eculicny (changes requested):
I think this looks good, but as we discussed and mentioned could use some other tests without changing the default casing for the existing tests.

It's a little noisy from the additional casing changes in the refactoring PR, but once that's merged they can be cleared up.

snowshu/adapters/base_sql_adapter.py
```diff
@@ -238,15 +239,18 @@ def quoted(val: str) -> str:
...
                         (case_corrected_database, 
                          case_corrected_schema, 
                          case_corrected_name)])
 
-    def _correct_case(self, val: str) -> str:
+    def _correct_case(self, val: str, attr: str = None) -> str:  # noqa pylint: disable=too-many-function-args
         """The base case correction method for a sql adapter.
         """
-        return val if self.preserve_case else correct_case(val, self.DEFAULT_CASE == 'upper')
+        preserve_case = self.preserve_case[attr] if attr in ['database', 'schema', 'relation'] else False
```
eculicny, line 253: We should limit the values passed to this function to be those in this list. Currently, any arbitrary string can be passed in and assumed to be `False`

PR #106, opened 2022-03-08
Remove testpypi publish step

eculicny:
Forgot that we can't overwrite file names in PyPi. Removed step to test that on merge into master.

Master pipeline now just builds the packaged versions of snowshu that would get deployed.


fleischer89 (approved):

.github/workflows/test_deployment.yml
```diff
@@ -25,4 +25,4 @@ jobs:
     - name: Build the Docker image
       run: docker-compose up -d
     - name: Creating new release package
-      run: docker-compose exec -T snowshu python3 setup.py sdist bdist_wheel
```
Aesthet, line 28: can't see any difference besides delete or added new line

eculicny: Ah, this change was also included in #103 which ended up getting merged before this PR. Missed that fact, going to close this one out.

eculicny:
Relevant changes added as part of #103 . Closed

PR #107, opened 2022-03-10
DATA-2303 removed creation of double fdw schemas in all DBs

glysav:


eculicny (commented):
The only additional thing that would be nice to have is some sort of test to ensure that the fix remains in place.

I don't think we need to create an entire additional test, but we can probably add some logic in [this test](https://github.com/Health-Union/snowshu/blob/master/tests/integration/snowflake/test_end_to_end.py#L360) to verify that no double FDW schemas exist.

eculicny (approved):

Aesthet (approved):

ArtemHU (approved):

PR #108, opened 2022-03-10
DATA-1959: Added support for `include_outliers` for polymorphic relationships

fleischer89:


eculicny (changes requested):
Definitely need some tests for what SQL the include outliers functionality is actually generating because I think that we're going to need two more constraint statements instead of one due to the differing behavior of the upstream vs downstream relations in the polymorphic dependencies.

Also, the slight differences between "child ids in parent table" and "parent ids in child tables" might result in slightly different constructions of the existing statements. Will give it some more thought, but tests will clear this up.

PR #110, opened 2022-04-20
Feature/data 2193

Gal40n04ek:


docs/source/user_documentation/user_documentation.rst
```diff
@@ -125,6 +125,25 @@ SnowShu will report details of the created replica once completed.
...
+
+>>> docker images
+
+It usually starts with ``snowshu_replica_``. So for a replica named *hamburger-sandwich*:
+
+>>> snowshu create -i snowshu_replica_hamburger-sandwich
+
+or
+
+>>> snowshu create --incremental snowshu_replica_hamburger-sandwich
```
eculicny, line 143: Do the images need to include the tag for this option? `snowshu_replica_hamburger-sandwich:latest`

Or does it default to `latest` when no version is present? I vaguely remember it requiring the version. If so we should document it either way (i.e. defaults to `latest` or make sure the version is included)

Gal40n04ek: Hello! Fixed. Please, check!

eculicny (approved):

Aesthet (approved):

PR #111, opened 2022-04-22
Feature/data 2477: integration tests for snowflake adapter

Gal40n04ek:


tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -0,0 +1,349 @@
...
+    statement = sf_adapter.population_count_statement(relation)
+
+    assert query_equalize(statement) == query_equalize(
+        f"SELECT COUNT(*) FROM {sf_adapter.quoted_dot_notation(relation)}")
+
+
+def test_get_all_databases(sf_adapter):
+    databases_list = ["SNOWSHU_DEVELOPMENT", "HU_DATA", "SNOWFLAKE_SAMPLE_DATA", "DEMO_DB", "UTIL_DB"]
+
+    assert databases_list.sort() == sf_adapter._get_all_databases().sort()
```
eculicny, line 88: For simplicity we can probably limit this to only `SNOWSHU_DEVELOPMENT` for now since the `snowshu_development_role` should only have access to that database, or do we need to adjust the permissions for it?

tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -0,0 +1,349 @@
...
+from tests.common import query_equalize, rand_string
+
+
+@pytest.fixture(scope='session')
+def sf_adapter():
+    with open(CREDENTIALS) as cred_file:
+        credentials = yaml.safe_load(cred_file)
+
+    profile_dict = get_connection_profile(credentials)
+    profile_dict.update({"role": "snowshu_development_role"})
```
eculicny, line 24: It's probably worth enforcing the role value to be present in the credentials file.

eculicny (commented):
Unless I'm missing something, some of these tests should probably be moved to the unit tests for the snowflake adapters since they don't actually rely on a connection to snowflake.

Some of the query comparison tests could be slightly altered to issue the queries against snowflake to ensure that we are generating valid SQL, especially since I think some of those comparison tests may already be covered elsewhere.

tests/integration/snowflake/test_snowflake_adapter.py
Aesthet: In that case it may be worth mentioning somewhere that: 

> You should use snowshu_development_user/role when running integration tests

tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -0,0 +1,347 @@
...
+    assert f" ({predicate} AND LOWER({LOCAL_TYPE}) = LOWER('{TYPE_MATCH_VAL}') ) " == \
+           sf_adapter.polymorphic_constraint_statement(relation,
+                                                       True,
+                                                       LOCAL_KEY,
+                                                       REMOTE_KEY,
+                                                       LOCAL_TYPE,
+                                                       TYPE_MATCH_VAL)
+
+
+def test_build_conn_string(sf_adapter):
```
eculicny, line 205: This one could probably be moved to the unit tests file as well or merged with the existing `test_conn_string_basic` test there by adding a randomly generated `ROLE` variable.

tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -0,0 +1,347 @@
...
+def test_get_connection(sf_adapter):
+    conn_string = sf_adapter.get_connection()
+    DATABASE = sf_adapter.credentials.database
+    USER = sf_adapter.credentials.user
+    PASSWORD = sf_adapter.credentials.password
+    ACCOUNT = sf_adapter.credentials.account
+    ROLE = sf_adapter.credentials.role
+
+    assert conn_string.url.render_as_string(hide_password=False) == \
+           f'snowflake://{USER}:{PASSWORD.replace("@", urllib.parse.quote_plus("@"))}@{ACCOUNT}/{DATABASE}/?role={ROLE}'
```
eculicny, line 338: This test might also be worth moving to unit tests with the other connection string test. Unless I'm missing something about this that actually calls out to snowflake.

tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -0,0 +1,347 @@
...
+            *
+        FROM 
+            {DATABASE}.{SCHEMA}.{TABLE}
+            SAMPLE BERNOULLI (10)
+        """)
+
+    assert len(sf_adapter._safe_query(sample)) > 0
+
+
+def test_analyze_wrap_statement(sf_adapter):
```
eculicny, line 257: This test should be able to be issued against snowflake since it's a fully formed query.

tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -0,0 +1,347 @@
...
+    sampling = BernoulliSampleMethod(50, units='probability')
+    query = """SELECT * FROM "SNOWSHU_DEVELOPMENT"."EXTERNAL_DATA"."ADDRESS_REGION_ATTRIBUTES"
+            WHERE IS_CURRENTLY_TARGETED = TRUE
+            AND SALES_REGION IN ('northeast', 'southeast')
+            AND PRIMARY_REGIONAL_CREDIT_PROVIDER = 'mastercard'"""
+    DATABASE, SCHEMA, TABLE = "SNOWSHU_DEVELOPMENT", "EXTERNAL_DATA", "ADDRESS_REGION_ATTRIBUTES"
+    relation = Relation(database=DATABASE,
+                        schema=SCHEMA,
+                        name=TABLE,
+                        materialization=TABLE,
```
eculicny, line 40: Sorry I missed this on the first pass, but the `materialization` attribute should be one of the materialization types created [here](https://github.com/Health-Union/snowshu/blob/ea0bf0eaa82a3d2834262025f19d3da6ab6027e4/snowshu/core/models/materializations.py#L5).

However, it might not be required for most of the relations here and could be left out entirely.

eculicny:
Actually, maybe one minor issue. Does running `pytest` work for you? It doesn't handle the same named file well even though they are in different subdirectories with the current structure.

Having the same files name in both `tests/integration` and `tests/unit` would probably work if we had `__init__.py` files in those directories based on pytest [docs](https://docs.pytest.org/en/stable/explanation/goodpractices.html#tests-outside-application-code). If not, renaming the new file will fix the problem.

tests/integration/snowflake/test_snowflake_adapter.py
Aesthet: Added by Halyna on line 284. Also I believe this also covers your comment "Looks good! The only outstanding thing I saw was one of the comments on test_analyze_wrap_statement, which currently isn't issued against snowflake, but should be for the integration test."?

Aesthet (approved):

tests/integration/snowflake/test_snowflake_adapter.py
eculicny: Yep.

PR #112, opened 2022-05-05
Feature/data 2478: integration tests for the postgres target adapter

Gal40n04ek:


tests/integration/postgres/test_postgres_target_adapter.py
```diff
@@ -0,0 +1,132 @@
...
+from snowshu.core.main import cli
+from snowshu.core.models import Relation, Attribute, data_types
+from snowshu.core.models.materializations import TABLE
+from tests.common import rand_string
+
+CONFIGURATION_PATH = os.path.join(PACKAGE_ROOT, 'tests', 'assets', 'replica_test_config.yml')
+DOCKER_SPIN_UP_TIMEOUT = 15
+
+@pytest.fixture(scope="module", autouse=True)
+def end_to_end(docker_flush_session):
```
eculicny, line 21: For the tests that use a similar replica to the one that was built for the end to end tests here, we might be able to extract that fixture so that both tests use the same test replica. 

This should save a decent chunk of time since we won't need to build it twice.

Gal40n04ek: Moved the fixture to the contftest.py file, but now the fixture has scope 'module', not 'session'. As I can see, there's another fixture (docker_flush) in the conftest file used for other tests and it has the function scope. This fixture has sanitize_docker_environment() method that removes containers and images. So I didn't find another solution for this, but to set a module scope. May be, there's a better way to solve my problem

eculicny: I think that should be fine to switch that to a module scope from session.

However, we should probably rename the fixture from `docker_flush_session` to `docker_flush_module` to keep in line with the naming convention.

eculicny (approved):

Aesthet (approved):

PR #113, opened 2022-05-10
feature/DATA-2471: add useful output for cycles in the dependency graph

Gal40n04ek:
@eculicny, @Aesthet, @ArtemHU Is it better to output the list of the nodes in the cycle or one node is enough? I add the list of the nodes in the logger message, but only one node in the Error message. Is it OK?


Aesthet:
> @eculicny, @Aesthet, @ArtemHU Is it better to output the list of the nodes in the cycle or one node is enough? I add the list of the nodes in the logger message, but only one node in the Error message. Is it OK?
I guess it is OK just to state the fact that there were cycles in exception, while stating all nodes in log messages. Other idea I personally had when we were pointing this task - it would be nice to visualise  those cycles either via log messages (would be difficult to visualise big/complicated cycles) or via some lib. If this will be to complicated or will require big libs - we can skip that, but at least try to dig a bit in that direction.

Aesthet:
I left few comments in your personal repo by mistake - please check those

snowshu/core/graph.py
```diff
@@ -91,8 +91,22 @@ def build_graph(self, configs: Configuration) -> None:
...
+                    continue
+                cycle_names.append(cycle_name)
+
+            logger.error(
+                f'The dependency graph generated by the given specified relations yields a cyclic graph. \
+                \n\tCyclic dependency found in the following relations:\n\t{graph_cycles}')
             raise ValueError(
-                'The graph created by the specified trail path is not directed (circular reference detected).')
+                f'The graph created by the specified trail path is not directed (circular reference detected): \
+                \n{cycle_names}')
```
eculicny, line 109: I think this was supposed to be included on the other branch/PR :)

eculicny:
I think we should try to dump more than just one node of one of the cycles since it's not very clear where the cycles were generated if the replica.yml file isn't well organized for it.

Networkx already has some features for drawing and dumping adjacency lists:
https://networkx.org/documentation/stable/reference/drawing.html#
https://networkx.org/documentation/stable/reference/readwrite/index.html

Depending on complicated we want to get, we could tie in the `--barf` option to the output. Without the flag only dump the first _n_ vertices of each cycle, with the `--barf` flag, dump the adjacency lists or drawings into the output directory? That might be overkill, and the adjacency lists in the CLI might work well enough.

snowshu/core/graph.py
```diff
@@ -91,6 +93,46 @@ def build_graph(self, configs: Configuration) -> None:
             f'Identified a total of {len(self.graph)} relations to sample based on the specified configurations.')
 
         if not networkx.algorithms.is_directed_acyclic_graph(self.graph):
+            graph_cycles = list(networkx.simple_cycles(self.graph))
```
Aesthet, line 96: This method have become really huge - maybe put some (or even all) logic related to presence of graph_cycles into separate method?

snowshu/core/graph.py
```diff
@@ -91,6 +93,46 @@ def build_graph(self, configs: Configuration) -> None:
...
+            plt.figure(figsize=(8, 8))
+            plt.margins(0.1)
+            plt.title(f'\nGraph of the simple cycles, created at: {created_at.strftime("%d/%m/%Y %H:%M:%S")}')
+
+            networkx.draw(
+                cycle_graph, pos=networkx.planar_layout(cycle_graph), labels=label_dict, with_labels=True,
+                node_size=1000, node_color='skyblue', font_size=6, font_color='green', width=2,
+                horizontalalignment='center', verticalalignment='center', connectionstyle='arc3, rad=0.05')
+
+            filename = f'snowshu_barf_output/{created_at.strftime("%d_%m_%Y_%H_%M_%S")}.'
```
Aesthet, line 127: Use GraphSetRunner.barf_output variable here instead of hard coded value

tests/integration/test_graph.py
```diff
@@ -0,0 +1,21 @@
...
+
+import pytest
+import yaml
+
+from snowshu.core.configuration_parser import ConfigurationParser
+from snowshu.core.graph import SnowShuGraph
+from tests.conftest import CYCLE_CONFIGURATION
+
+
+def test_build_graph_cycle_output():
```
eculicny, line 12: You probably could have left this in the unit tests.

snowshu/core/graph.py
```diff
@@ -70,6 +69,58 @@ def catalog_difference(source_graph: Union["SnowShuGraph", networkx.Graph],
...
+        plt.figure(figsize=(8, 8))
+        plt.margins(0.1)
+        plt.title(f'\nGraph of the simple cycles, created at: {created_at.strftime("%d/%m/%Y %H:%M:%S")}')
+
+        networkx.draw(
+            cycle_graph, pos=networkx.planar_layout(cycle_graph), labels=label_dict, with_labels=True,
+            node_size=1000, node_color='skyblue', font_size=6, font_color='green', width=2,
+            horizontalalignment='center', verticalalignment='center', connectionstyle='arc3, rad=0.05')
+
+        filename = f'{GraphSetRunner.barf_output}/{created_at.strftime("%d_%m_%Y_%H_%M_%S")}'
```
eculicny, line 118: We should probably name the files appropriately instead of just using a datetime.

Nitpick: The timestamp should also be reordered year-month-day instead of day-month-year. Same for the graph title. It should hopefully avoid any confusion on timestamp format.

eculicny (approved):
I think this looks good.

We might want to revisit where the error handling is happening at some point in the future. As in moving the building of the cycle output and image up one layer to `graph_set_runner` which would be a better location so we can easily make use of the `barf` boolean there.

snowshu/core/graph.py
```diff
@@ -66,6 +70,62 @@ def catalog_difference(source_graph: Union["SnowShuGraph", networkx.Graph],
         source_graph.remove_nodes_from(nodes_to_delete)
         return source_graph
 
+    @staticmethod
+    def _build_graph_cycles_output(graph_cycles: list) -> Tuple[str, str]:
+        """ Builds simple cycles in a directed graph per replica config.
+
+            Args:
+                configs: :class:`Configuration <snowshu.core.configuration_parser.Configuration>` object.
```
ArtemHU, line 78: the input parameter is defined as list
should we change the description?

snowshu/core/graph.py
```diff
@@ -66,6 +70,62 @@ def catalog_difference(source_graph: Union["SnowShuGraph", networkx.Graph],
...
+                nodes.append(node)
+                if i != len(cycle) - 1:
+                    cycle_graph.add_edge(node, cycle[i + 1])
+                    message = message + '\033[1;34m' + str(node)[17:-1] + '\t\033[1;32m----\t'
+                else:
+                    cycle_graph.add_edge(node, cycle[0])
+                    message = message + '\033[1;34m' + str(node)[17:-1]
+            message = message + '\n\t'
+
+        filename = ''
```
ArtemHU, line 102: looks like it is defined in advance (I mean the filename variable)

snowshu/core/graph.py
```diff
@@ -91,6 +151,20 @@ def build_graph(self, configs: Configuration) -> None:
...
         if not networkx.algorithms.is_directed_acyclic_graph(self.graph):
+            graph_cycles = list(networkx.simple_cycles(self.graph))
+            message, filename = self._build_graph_cycles_output(graph_cycles)
+
+            if filename:
+                logger.error(
+                    'The dependency graph generated by the given specified relations yields a cyclic graph. \
+                    \n\tCyclic dependency found in the following relations:\n\t%s \
+                    \n\t\033[1;37mThe network topology diagram has been saved to: \
+                    \n\t\033[0;36m%s.png \033[1;37m and \033[0;36m%s.graphml', message, filename, filename)
```
ArtemHU, line 162: why don't to use formatting or f-string

PR #115, opened 2022-05-17
Snowshu: Add configurable MAXALLOWED_ROWS to replica.yml

skavierin:
https://health-union.atlassian.net/browse/DATA-2569

### DIFF
--- a/tests/unit/test_graph_runner.py
+++ b/tests/unit/test_graph_runner.py
@@ -3,6 +3,7 @@ from unittest import mock
 
 import pandas as pd
 import networkx as nx
+from requests import patch
 
 from snowshu.core.graph_set_runner import GraphExecutable, GraphSetRunner
 from snowshu.samplings.samplings import DefaultSampling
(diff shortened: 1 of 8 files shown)


eculicny (commented):
I think this looks good.

Can we add a small note or example in the documentation about the location in the YAML config for it? I think [here](https://github.com/Health-Union/snowshu/blob/master/docs/source/user_documentation/replica_dot_yaml_file.rst#L89-L93) may be the best place since we don't fully deconstruct the yaml in the docs.

tests/unit/test_graph_runner.py
```diff
@@ -53,3 +54,46 @@ def test_traverse_and_execute_analyze(stub_graph_set):
...
+    source_adapter.check_count_and_query.return_value=pd.DataFrame([dict(population_size=1000,sample_size=100)])
+    dag=copy.deepcopy(graph_set[-1]) # last graph in the set is the dag
+    
+    # test if defaults are passed
+    for rel in dag.nodes:
+        rel.unsampled=False
+        rel.include_outliers=False
+        rel.sampling=DefaultSampling()
+
+    dag_executable = GraphExecutable(dag, source_adapter, target_adapter, True)
```
ArtemHU, line 82: it looks like in both cases the analyse attribute of the GraphExecutable is defined as True;
Meantime, there is uncovered call to 'check_count_and_query' when the analyse attribute eq to False
for more details please check the `_traverse_and_execute` function of the GraphSetRunner class

ArtemHU:
> I think this looks good.
> 
> Can we add a small note or example in the documentation about the location in the YAML config for it? I think [here](https://github.com/Health-Union/snowshu/blob/master/docs/source/user_documentation/replica_dot_yaml_file.rst#L89-L93) may be the best place since we don't fully deconstruct the yaml in the docs.

Completely agree, we should review and extend the existing documentation.
What do you think about creating an additional task(separate jira or sub task) to update the rst doc?
cc @eculicny

tests/unit/test_graph_runner.py
skavierin: Done

skavierin:
> > > I think this looks good.
> > > Can we add a small note or example in the documentation about the location in the YAML config for it? I think [here](https://github.com/Health-Union/snowshu/blob/master/docs/source/user_documentation/replica_dot_yaml_file.rst#L89-L93) may be the best place since we don't fully deconstruct the yaml in the docs.
> > 
> > 
> > Completely agree, we should review and extend the existing documentation. What do you think about creating an additional task(separate jira or sub task) to update the rst doc? cc @eculicny
> 
> I'd like to have the example included as part of this PR just to indicate where the new option is supposed to fall in the yaml structure. All that would need to be is adding the new `max_allowed_rows` option to the example `sampling` override block I linked in that comment.
> 
> Overall the documentation could use an overhaul, and the replica yaml could do with a better (and probably partitioned out) explanation which we can include in some other ticketing work. That would be where we can fully explain all of the options and configs available.

Done

eculicny (approved):

ArtemHU (approved):

PR #117, opened 2022-05-26
Feature/data 2557

skavierin:


tests/unit/test_init.py
```diff
@@ -86,3 +90,55 @@ def test_analyze_does_all_but_run(replica, create_relation):
...
         replica.assert_called_once()
         create_relation.assert_not_called()
+
+
+@patch('snowshu.core.main.ReplicaFactory.create')
+def test_custom_retry_count_cli_input(create):
+    # test if CLI input is passed to replica.create
+    runner = CliRunner()
+    runner.invoke(main.cli, ('create'))
+    create.assert_called_with(name=ANY, barf=ANY, retry_count=1)
```
skavierin, line 100: This bit passes locally but fails in CI, needs investigation

ArtemHU: I tested it locally - it fails
image

Could you please run all unit test cases from docker?
Also, please check the version of pytest

skavierin: Ran as it is supposed to be done 
image

ArtemHU: I reason why it fails - you have no replica.yml file in your root folder
By default it should reside in the root folder

ArtemHU: it looks like you need to supply the replica.yml file

skavierin: Guess it's fixed, recent run succeeded

snowshu/core/replica/replica_factory.py
```diff
@@ -68,6 +73,7 @@ def _execute(self,
                                  self.config.source_profile.adapter,
                                  self.config.target_profile.adapter,
                                  threads=self.config.threads,
+                                 retry_count=self.retry_count,
```
ArtemHU, line 76: should we add it to cli?

eculicny: Let's add it in to the `analyze` command for consistency between available options.

eculicny (commented):
Looks good! Only need to have the extra argument propagated to the `test_analyze.py` integration test.

skavierin:
> Looks good! Only need to have the extra argument propagated to the test_analyze.py integration test.

Done

ArtemHU (approved):

PR #119, opened 2022-05-31
Feature/data 2703

ArtemHU:


snowshu/core/graph.py
```diff
@@ -365,8 +367,26 @@ def _process_downstream_relation_set(
...
             for upstream_relation in upstream_without_downstream:
+                # validate whether relation is valid
+                attributes = dict(filter(lambda item: item[0] in ['remote_attribute', 'local_attribute'],
+                                         relationship['edge_attributes'].items()))
+                remote_attribute = attributes.get('remote_attribute')
+                local_attribute = attributes.get('local_attribute')
+                is_remote_attribute_valid = remote_attribute in catalog_dict.get(upstream_relation.dot_notation, {})
+                is_local_attribute_valid = local_attribute in catalog_dict.get(downstream_relation.dot_notation, {})
+
+                if not is_remote_attribute_valid or not is_local_attribute_valid:
```
ArtemHU, line 383: Maybe, I should accumulate all possible issues and only then run raise an issue?
cc @eculicny

eculicny: It would probably be best to do that. Then the user doesn't need to keep trying to run when there are multiple problems that can be identified on one validation.

ArtemHU: image

(i) It will accumulate relevant information for every single db

eculicny (approved):

glysav (approved):

PR #122, opened 2022-06-17
DATA-2806: the initial commit, the replica data share

ArtemHU:


Dockerfile
```diff
@@ -1,10 +1,10 @@
 ## docker image is intended for development environments only.
 ## to use snowshu in production install the package via pip
-FROM python:3.8-slim-buster
+FROM python:3.9-slim-buster
 
-RUN mkdir /app
```
eculicny, line 5: Why switch to `/tmp/app` here?

ArtemHU: the /tmp folder is added to the "file sharing" by default
this is needed to support the run from container

requirements/base.txt
```diff
@@ -6,7 +6,7 @@ coloredlogs==15.0
 networkx==2.6.2
 tabulate==0.8.8
 psycopg2-binary==2.9.1
-scipy==1.7.3
+scipy==1.8.1
```
ArtemHU, line 9: a good instruction on scipy + m1:https://stackoverflow.com/questions/65745683/how-to-install-scipy-on-apple-silicon-arm-m1

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -75,11 +77,13 @@ def __init__(self, replica_metadata: dict, **kwargs):
...
         self.x00_replacement = kwargs.get("pg_0x00_replacement", "")
 
-        self.DOCKER_START_COMMAND = f'postgres -p {self._credentials.port} ' # noqa pylint: disable=invalid-name
-        self.DOCKER_READY_COMMAND = (f'pg_isready -p {self._credentials.port} ' # noqa pylint: disable=invalid-name
+        self.DOCKER_START_COMMAND = f'postgres -p {self._credentials.port} '  # noqa pylint: disable=invalid-name
+        self.DOCKER_READY_COMMAND = (f'pg_isready -p {self._credentials.port} '  # noqa pylint: disable=invalid-name
                                      f'-h {self._credentials.host} '
                                      f'-U {self._credentials.user} '
                                      f'-d {self._credentials.database}')
+        self.DOCKER_SHARE_REPLICA_DATA = f"mkdir {self.DOCKER_REPLICA_MOUNT_FOLDER} || " \
```
eculicny, line 85: Do we need the `mkdir` step here if we're assuming the volume has already been mounted when the container starts?

snowshu/core/replica/replica_factory.py
```diff
@@ -100,6 +102,14 @@ def _execute(self,
...
             logger.info('Emulation functions applied.')
+
+            logger.info('Copying replica data to shared location...')
+            status_message = self.config.target_profile.adapter.copy_replica_data()
+            if status_message[0] != 0:
+                message = (f'Failed to execute copy command: {status_message[1]}')
+                logger.error(message)
+                raise UnableToExecuteCopyReplicaCommand(message)
+
             self.config.target_profile.adapter.finalize_replica()
```
eculicny, line 113: Do we want to clean up the docker volume once we're done so it isn't taking up space on the host system as part of this? Or leave that until later work when we know exactly where the volume will be used?

glysav (approved):

PR #123, opened 2022-06-22
DATA-2805 - feature/DATA-2805

ArtemHU:


snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -178,6 +178,17 @@ def initialize_replica(self,
     def _init_image(self,
                     source_adapter_name: str) -> None:
         shdocker = SnowShuDocker()
+
+        image_architecture = shdocker.get_docker_image_attributes(self.DOCKER_IMAGE) \
+            .get('Architecture', LOCAL_ARCHITECTURE)
```
eculicny, line 183: What case does defaulting to the `LOCAL_ARCHITECTURE` value support? Defaulting to that value ensures that a missing `"Architecture"` key results in no mismatch between image and host architectures.

snowshu/configs.py
```diff
@@ -34,6 +36,17 @@ def _is_in_docker() -> bool:
         return False
 
 
+def _get_architecture() -> str:
+    """
+     Returns the machine type. An empty string is returned if the value cannot be determined.
+    """
+    iso_arch = platform.machine()
```
eculicny, line 43: What does the Apple M1 return for this value?

Some quick testing on my side ended up with both in and out of a docker image returning `x86_64`

snowshu/adapters/target_adapters/base_target_adapter.py
ArtemHU: let's put the 'Unknown' as the default - anyway we treat it as a "warning" message

snowshu/configs.py
ArtemHU: 1) local run:
image
2) docker run(the arm image):
image

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -169,9 +175,20 @@ def initialize_replica(self,
                 raise error
         self._init_image(source_adapter_name)
 
-    def _init_image(self, 
+    def _init_image(self,
                     source_adapter_name: str) -> None:
         shdocker = SnowShuDocker()
+
+        image_architecture = shdocker.get_docker_image_attributes(self.DOCKER_IMAGE) \
+            .get('Architecture', LOCAL_ARCHITECTURE)
```
glysav: Is there are specific case when image attribute "Architecture" has no value, so we snould to specify `LOCAL_ARCHITECTURE` as default value?

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -311,6 +312,10 @@ def statement_runner(statement: str):
                     statement_runner(f'IMPORT FOREIGN SCHEMA {schema} FROM SERVER '
                                      f'{schema_database} INTO {schema_database}__{schema}')
 
+    def copy_replica_data(self) -> Tuple[bool, str]:
+        status = self.container.exec_run(f"/bin/bash -c '{self.DOCKER_SHARE_REPLICA_DATA}'", tty=True)
```
glysav, line 315: we can return status without assigning variable

snowshu/configs.py
ArtemHU: we are getting the following warning(in case of arch mismatch):
image

glysav (approved):

eculicny (approved):

PR #124, opened 2022-06-30
Data 2472

skavierin:


.gitignore
```diff
@@ -14,3 +14,5 @@ keyfile.json
 
 tests/assets/integration/trail-path.yml
 tests/assets/integration/credentials.yml
+credentials.yml
+replica.yml
```
eculicny, line 18: These two files are used by the init/new command, so should still be included. We'd need to rename some things if we want to always exclude generic replica/credentials files.

eculicny (commented):
Looks good. Only one minor change needed with the ignored files.

.gitignore
skavierin: On my end files in templates folder were not affected, I will double check this.

skavierin: Yep, only the actual custom files are affected, templates are safe. Or did I misunderstood why we need that?

ArtemHU (approved):

snowshu/core/utils.py
```diff
@@ -4,17 +4,16 @@
...
-from snowshu.logger import Logger
-
 if TYPE_CHECKING:
     from snowshu.adapters.base_sql_adapter import BaseSQLAdapter
     from snowshu.adapters.source_adapters.base_source_adapter import BaseSourceAdapter
     from snowshu.adapters.target_adapters.base_target_adapter import BaseTargetAdapter
 
 
-logger = Logger().logger
+logger = logging.getLogger(__name__)
```
ArtemHU, line 16: it will be represented as snowshu.adapters.<name> (the root logger is snowshu) - as expected

eculicny (approved):

PR #126, opened 2022-07-01
Data 2473

skavierin:


snowshu/core/main.py
```diff
@@ -20,14 +20,21 @@
 
 
 @click.group()
-@click.option('--debug', '-d', is_flag=True, default=False,
-              help="run commands in debug mode")
-def cli(debug: bool):
+@click.option('--debug-core', '-v', is_flag=True, default=False, help='Set log level to debug only in core')
+@click.option('--debug-adapters', '-vv', is_flag=True, default=False, help='Set log level to debug everywhere')
+@click.option('--debug', '-d', is_flag=True, default=False, help='Set log level to debug everywhere')
+def cli(debug: bool, debug_core: bool, debug_adapters: bool):
```
skavierin, line 26: I left -d is left for compatibility. Also had intention of making these exclusive so you cannot use several options but did not find a way to do this. If anyone knows how, I will adjust this.

eculicny: While looking at options for the mutually exclusive situation, I came across [this](https://click.palletsprojects.com/en/8.1.x/options/#counting) functionality in click for counting options.

We might be able to split the options out a bit more
* `-v...` - various levels of verbosity (as currently defined)
* `--debug-core` - turns on debugging in core module only
* `--debug-adapters` - turns on debugging in the adapter modules only
* `--debug`, `-d` - same as max verbosity

We probably don't need to worry about mutual exclusivity as long as it's made clear in the description that all of the debug options just compound on one another if multiple are present.

Relating to the original question, here are two packages for argument groups. Never used either one, but both seem to support mutually exclusive options without writing the logic ourselves:
* https://github.com/janluke/cloup
* https://github.com/click-contrib/click-option-group

eculicny (commented):
Looks good. Needs a minor change [here](https://github.com/Health-Union/snowshu/blob/master/tests/integration/test_docker_end_to_end.py#L13) since the `set_log_level` function was changed.

ArtemHU (approved):

eculicny (approved):

PR #127, opened 2022-08-19
DATA-2807

skavierin:
Done not according to ticket.
Flow as built here:
read config.py for desired architectures -> build one or two pg containers according to config -> populate main container -> commit both containers since they now contain identical data.

### DIFF
--- a/snowshu/adapters/target_adapters/base_target_adapter.py
+++ b/snowshu/adapters/target_adapters/base_target_adapter.py
@@ -185,7 +185,8 @@ AS
                 .get('Architecture', 'Unknown architecture')
 
             if image_architecture.lower() == LOCAL_ARCHITECTURE.lower():
-                logger.info(f'The architecture of the docker image[{self.DOCKER_IMAGE}] is {image_architecture.lower()}')
+                logger.info(
+                    f'The architecture of the docker image[{self.DOCKER_IMAGE}] is {image_architecture.lower()}')
             else:
                 logger.warning(f"""There is a mismatch between local and docker image architectures: '
                                     The local architecture is {LOCAL_ARCHITECTURE} 
--- a/snowshu/adapters/target_adapters/base_target_adapter.py
+++ b/snowshu/adapters/target_adapters/base_target_adapter.py
@@ -200,7 +200,8 @@ AS
             self.DOCKER_READY_COMMAND).exit_code == 0
 
     def finalize_replica(self) -> None:
-        """returns the image name of the completed replica.
+        """ Converts all containers to respective replicas, 
+            creates 'latest' if any on the running containers are of local arch
         """
         shdocker = SnowShuDocker()
         logger.info('Finalizing target container into replica...')
--- a/snowshu/adapters/target_adapters/base_target_adapter.py
+++ b/snowshu/adapters/target_adapters/base_target_adapter.py
@@ -3,14 +3,13 @@ from datetime import datetime
 from time import sleep
 from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple
 import logging
-import docker
 
 import pandas as pd
 
 from snowshu.adapters import BaseSQLAdapter
 from snowshu.configs import (DEFAULT_INSERT_CHUNK_SIZE,
                              DOCKER_TARGET_CONTAINER, DOCKER_TARGET_PORT,
-                             IS_IN_DOCKER, LOCAL_ARCHITECTURE)
+                             IS_IN_DOCKER)
 from snowshu.core.docker import SnowShuDocker
 from snowshu.core.models import Attribute, Credentials, Relation
 from snowshu.core.models import DataType, data_types as dt
@@ -180,7 +179,7 @@ AS
     def _init_image(self,
                     source_adapter_name: str) -> None:
         shdocker = SnowShuDocker()
-        
+
         logger.info('Initializing target container...')
         self.container, self.passive_container = shdocker.startup(
             self.DOCKER_IMAGE,
--- a/snowshu/adapters/target_adapters/base_target_adapter.py
+++ b/snowshu/adapters/target_adapters/base_target_adapter.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from time import sleep
 from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple
 import logging
+import docker
 
 import pandas as pd
 
@@ -43,6 +44,7 @@ class BaseTargetAdapter(BaseSQLAdapter):
         self.target = DOCKER_TARGET_CONTAINER if IS_IN_DOCKER else 'localhost'
         self.credentials = self._generate_credentials(self.target)
         self.container: "Container" = None
+        self.passive_container: "Container" = None
         self.replica_meta = replica_metadata
 
     def enable_cross_database(self) -> None:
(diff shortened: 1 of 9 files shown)


snowshu/core/docker.py
```diff
@@ -31,24 +31,28 @@ def _create_snowshu_volume(self, volume_name: str) -> docker.models.volumes.Volu
...
-            repository=self.sanitize_replica_name(replica_name)
-        )
-        logger.info(f'Replica image {replica.tags[0]} created. Cleaning up...')
-        self.remove_container(container.name)
-
-        return replica
+        container_list = [
+            active_container, passive_container] if passive_container else [active_container]
+        for container in container_list:
+            container.start()
```
eculicny, line 43: Why do we need to restart each container here?

snowshu/core/docker.py
```diff
@@ -31,24 +31,28 @@ def _create_snowshu_volume(self, volume_name: str) -> docker.models.volumes.Volu
     def convert_container_to_replica(
             self,
             replica_name: str,
-            container: docker.models.containers.Container) -> docker.models.images.Image:
+            active_container: docker.models.containers.Container,
+            passive_container: docker.models.containers.Container) -> None:
```
eculicny, line 35: We should probably return all of the images instead of None here with the change

eculicny (commented):
I need to spend some time testing it, but the changes look sound. Mostly just a lot of questions here and there about small things.

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -178,19 +180,22 @@ def initialize_replica(self,
...
+
+            if image_architecture.lower() == LOCAL_ARCHITECTURE.lower():
+                logger.info(
+                    f'The architecture of the docker image[{self.DOCKER_IMAGE}] is {image_architecture.lower()}')
+            else:
+                logger.warning(f"""There is a mismatch between local and docker image architectures: '
+                                    The local architecture is {LOCAL_ARCHITECTURE} 
+                                    The image architecture is [{self.DOCKER_IMAGE}]({image_architecture})""")
+        except docker.errors.ImageNotFound:
+            logger.info('No existing target image found, creating')
```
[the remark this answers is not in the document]
skavierin: Yeah, before I changed it it would crash if you don't have any postgres image locally

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -43,6 +44,7 @@ def __init__(self, replica_metadata: dict):
         self.target = DOCKER_TARGET_CONTAINER if IS_IN_DOCKER else 'localhost'
         self.credentials = self._generate_credentials(self.target)
         self.container: "Container" = None
+        self.passive_container: "Container" = None
```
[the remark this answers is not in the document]
skavierin: This might be reasonable, but I do not foresee any other architecture. We sure won't support in on 32 bit arm (raspberry pie for example)

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -313,6 +314,13 @@ def statement_runner(statement: str):
 
     def copy_replica_data(self) -> Tuple[bool, str]:
         status = self.container.exec_run(f"/bin/bash -c '{self.DOCKER_SHARE_REPLICA_DATA}'", tty=True)
+        if self.passive_container:
+            self.container.stop()
+            self.passive_container.start()
+            logger.info('Copying replica data into passive container')
+            self.passive_container.exec_run(f"/bin/bash -c '{self.DOCKER_IMPORT_REPLICA_DATA_FROM_SHARE}'", tty=True)
+            self.passive_container.stop()
+            self.container.start()
```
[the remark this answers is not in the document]
skavierin: We cannot have both online at the same time due to them occupying same ports, hence the start/stop in multiple places. We need to restart the main container because you cannot commit the stopped one (I guess, might not be true), and the code I made in general assumes that active is running and passive is stopped.

snowshu/core/docker.py
skavierin: Because active container is committed and removed by this point, and passive one is stopped so it cannot be committed if it is not running. Trying to start a running container does nothing and returns None.

snowshu/core/docker.py
skavierin: I have though about that, but the return is not used anywhere.

snowshu/adapters/target_adapters/base_target_adapter.py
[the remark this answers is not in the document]
skavierin: Actually, I think this entire block can be removed since I removed the image.get(image_name) call further down the line. If the images are cached, the pull command will be nearly instant and will guarantee we pull the image of correct arch, instead of whatever latest postgres:12 we have

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -208,15 +199,15 @@ def target_database_is_ready(self) -> bool:
         return self.container.exec_run(
             self.DOCKER_READY_COMMAND).exit_code == 0
 
-    def finalize_replica(self) -> str:
+    def finalize_replica(self) -> None:
         """returns the image name of the completed replica.
```
ArtemHU, line 203: should we update the docstring?

skavierin: Yep, will fix

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -83,6 +84,7 @@ def __init__(self, replica_metadata: dict, **kwargs):
                                      f'-U {self._credentials.user} '
                                      f'-d {self._credentials.database}')
         self.DOCKER_SHARE_REPLICA_DATA = f"cp -af $PGDATA/* {self.DOCKER_REPLICA_MOUNT_FOLDER}"  # noqa pylint: disable=invalid-name
+        self.DOCKER_IMPORT_REPLICA_DATA_FROM_SHARE = f"cp -r -f {self.DOCKER_REPLICA_MOUNT_FOLDER}/* $PGDATA/"  # noqa pylint: disable=invalid-name
```
ArtemHU, line 87: any reason for "-r"

skavierin: To copy items in subdirs as well

snowshu/adapters/target_adapters/base_target_adapter.py
skavierin: Actually, there was a return that I removed due to it not being used by anything, hence 'None" is valid

snowshu/core/docker.py
```diff
@@ -84,24 +98,61 @@ def get_stopped_container(  # noqa pylint: disable=too-many-arguments
...
-        logger.info(f"Created stopped container {container.name}.")
-        return container
+        if not TARGET_ARCHITECTURE:
+            arch_list = [LOCAL_ARCHITECTURE]
+        else:
+            arch_list = TARGET_ARCHITECTURE
+
+        logger.info(f'Finding base image {image_name}...')
+        container_list = []
+        for arch in arch_list:
```
ArtemHU, line 108: (?) for the multi-arch case, the image-pull command should be executed in parallel 
can be  improved later

skavierin: Images are pulled as postgres:12 and that exact image instance is retagged to represent its arch, furthermore you cannot start those in parallel due to same ports.

ArtemHU (approved):

eculicny (approved):

PR #128, opened 2022-09-05
Fixed bug with special symbols in password.

Aesthet:


snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -295,11 +296,10 @@ def quoted(val: str) -> str:
     @overrides
     def _build_conn_string(self, overrides: Optional[dict] = None) -> str:  # noqa pylint: disable=redefined-outer-name
         """overrides the base conn string."""
-        conn_parts = [f"snowflake://{self.credentials.user}:{self.credentials.password}"
-                      f"@{self.credentials.account}/{self.credentials.database}/"]
```
eculicny, line 299: Is it worth quoting almost all of them to avoid any similar future issues?

Aesthet: Potentially yes, was thinking about that as well

glysav:
I believe some tests should be updated correspondingly. At least [this one](https://github.com/Aesthet/snowshu/blob/bugfix/DATA-3322/tests/unit/test_snowflake_adapter.py#L46) shouldn't work I think

glysav (approved):

eculicny (approved):

PR #131, opened 2022-10-06
feature/DATA-3475

ArtemHU:
a fix to dataframe gc issue


eculicny (approved):

eculicny:
Actually, to avoid having the issue be silent in the future, can we add in an info log message [here](https://github.com/Health-Union/snowshu/blob/master/snowshu/core/graph_set_runner.py#L213-L214), so it isn't quite as silently behaving incorrectly.

ArtemHU:
> Actually, to avoid having the issue be silent in the future, can we add in an info log message [here](https://github.com/Health-Union/snowshu/blob/master/snowshu/core/graph_set_runner.py#L213-L214), so it isn't quite as silently behaving incorrectly.

yeah, it makes sense to me

glysav (approved):

PR #132, opened 2022-10-07
Data 3540

skavierin:
Changed CLI args to use multiarch feature. It is now called by `snowshu create -m`

### DIFF
--- a/snowshu/core/main.py
+++ b/snowshu/core/main.py
@@ -5,10 +5,8 @@ from shutil import copyfile, which
 
 import click
 
-from snowshu.configs import IS_IN_DOCKER, \
-                            DEFAULT_RETRY_COUNT,\
-                            LOCAL_ARCHITECTURE,\
-                            MULTIARCH_LIST
+from snowshu.configs import IS_IN_DOCKER, DEFAULT_RETRY_COUNT, LOCAL_ARCHITECTURE
+from snowshu.core.utils import get_multiarch_list
 from snowshu.core.replica.replica_factory import ReplicaFactory
 from snowshu.core.replica.replica_manager import ReplicaManager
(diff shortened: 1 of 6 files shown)


snowshu/core/main.py
```diff
@@ -116,28 +117,22 @@ def source(filename):
     default=DEFAULT_RETRY_COUNT
 )
 @click.option(
-    '--architecture', '-arch',
-    help="Select arch for replicas, use as 'snowshu create -arch arm64 -arch amd64'",
-    default=None,
-    multiple=True
+    '--multiarch', '-m',
+    help="Tells SnowShu to build replicas of all architectures",
```
eculicny, line 121: We should probably keep the list of architectures in the help bit for reference since it isn't a long list currently.

eculicny (approved):

tests/unit/test_init.py
```diff
@@ -131,21 +131,15 @@ def test_custom_cli_input_load(load, temporary_replica):  # noqa pylint: disable
...
-
-    runner.invoke(main.cli, ('create -arch arm64'))
-    load.assert_called_with(ANY, target_arch=['arm64'])
-
-    runner.invoke(main.cli, ('create -arch amd64 -arch arm64'))
-    load.assert_called_with(ANY, target_arch=['amd64', 'arm64'])
-
-    with pytest.raises(ValueError):
-        runner.invoke(main.cli, ('create -arch incorrect-arch'), catch_exceptions=False)
+    for local_arch in ['arm64', 'amd64']:
```
ArtemHU, line 134: potentially we can leverage the "@pytest.mark.parametrize" here

ArtemHU (approved):

PR #133, opened 2022-10-11
DATA-3542: Fix incremental multiarch builds

skavierin:
- Core functionality done as described in a ticket
- Reference to `postgres:12` moved to `configs.py` for use in the future

### DIFF
--- a/snowshu/configs.py
+++ b/snowshu/configs.py
@@ -17,6 +17,7 @@ DOCKER_TARGET_CONTAINER = 'snowshu_target'
 DOCKER_REMOUNT_DIRECTORY = 'snowshu_replica_data'
 DOCKER_TARGET_PORT = 9999
 DOCKER_WORKING_DIR = Path('/app').as_posix()
+POSTGRES_IMAGE = 'postgres:12'
 
 
 def _is_in_docker() -> bool:
--- a/snowshu/core/docker.py
+++ b/snowshu/core/docker.py
@@ -7,7 +7,7 @@ import logging
 import docker
 
 from snowshu.configs import (DOCKER_NETWORK, DOCKER_TARGET_CONTAINER, DOCKER_REPLICA_MOUNT_FOLDER,
-                             DOCKER_WORKING_DIR, DOCKER_REPLICA_VOLUME, LOCAL_ARCHITECTURE)
+                             DOCKER_WORKING_DIR, DOCKER_REPLICA_VOLUME, LOCAL_ARCHITECTURE, POSTGRES_IMAGE)
 
 if TYPE_CHECKING:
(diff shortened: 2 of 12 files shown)


snowshu/core/docker.py
```diff
@@ -82,6 +82,7 @@ def convert_container_to_replica(
     def get_stopped_container(  # noqa pylint: disable=too-many-arguments
```
eculicny, line 82: If we're going to be referencing and working with the expected image for the target adapter, we should start passing in the target adapter as an argument instead of directly referencing the image via `POSTGRES_IMAGE`.

The functionality here in `docker.py` should essentially be image agnostic with the exception of some of the architecture verification.

eculicny (changes requested):
Unless I missed something, the logic looks sound, but we should try to maintain the existing abstractions instead of coupling the code to the assumption of a postgres target.

snowshu/core/docker.py
skavierin: Done

snowshu/configs.py
```diff
@@ -17,6 +17,7 @@
 DOCKER_REMOUNT_DIRECTORY = 'snowshu_replica_data'
 DOCKER_TARGET_PORT = 9999
 DOCKER_WORKING_DIR = Path('/app').as_posix()
+POSTGRES_IMAGE = 'postgres:12'
```
eculicny, line 20: Since we don't need this anywhere else, it might be worth leaving the postgres image in that specific adapter

eculicny (approved):

snowshu/configs.py
skavierin: That is used in some integration tests, so I think it's better this way than engineering a way to correctly inject target adapter everywhere.

tests/integration/snowflake/test_incremental_build.py
```diff
@@ -0,0 +1,152 @@
...
+                                                  detach=True)
+                time.sleep(DOCKER_SPIN_UP_TIMEOUT)
+
+                assert get_tables(
+                    'integration-test') == expected_incremented_table_list
+                container.remove(force=True)
+
+            for image in client.images.list('snowshu_replica_integration-test'):
+                for tag in image.tags:
+                    client.images.remove(tag)
```
ArtemHU, line 152: it looks to me that this test is overloaded with logic, why don't to create a set of small tests?
benefit: easier to maintain

skavierin: It's all dependent on the order of execution. I could split it for regular and multiarch variants, but not much after that.

ArtemHU (approved):

PR #134, opened 2022-10-18
hotfix/DATA-3542

skavierin:
- Fixes to incremental build logic. Should now actually work as intended.
- docker.py cleanup, reduced branching and arg list for `.startup()` since almost all of the info is already within `target_adapter`

### DIFF
--- a/snowshu/adapters/target_adapters/base_target_adapter.py
+++ b/snowshu/adapters/target_adapters/base_target_adapter.py
@@ -10,6 +10,7 @@ from snowshu.adapters import BaseSQLAdapter
 from snowshu.configs import (DEFAULT_INSERT_CHUNK_SIZE,
                              DOCKER_TARGET_CONTAINER, DOCKER_TARGET_PORT,
                              IS_IN_DOCKER)
+from snowshu.core.docker import SnowShuDocker
 from snowshu.core.models import Attribute, Credentials, Relation
 from snowshu.core.models import DataType, data_types as dt
 from snowshu.core.models import materializations as mz
@@ -23,7 +24,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-class BaseTargetAdapter(BaseSQLAdapter):
+class BaseTargetAdapter(BaseSQLAdapter):  # noqa pylint: disable=too-many-instance-attributes
     """All target adapters inherit from this one."""
     REQUIRED_CREDENTIALS = [USER, PASSWORD, HOST, PORT, DATABASE]
     ALLOWED_CREDENTIALS = []
@@ -41,9 +42,10 @@ class BaseTargetAdapter(BaseSQLAdapter):
 
         self.target = DOCKER_TARGET_CONTAINER if IS_IN_DOCKER else 'localhost'
         self.credentials = self._generate_credentials(self.target)
+        self.shdocker = SnowShuDocker()
         self.container: "Container" = None
         self.passive_container: "Container" = None
-        self.shdocker = None
+        self.target_arch = None
         self.replica_meta = replica_metadata
         self.is_incremental = False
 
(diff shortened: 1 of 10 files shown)


snowshu/core/docker.py
```diff
@@ -144,20 +116,80 @@ def get_stopped_container(  # noqa pylint: disable=too-many-arguments
...
+                                         'already allocated, stop extra containers and rerun')
+                    raise
+
+                logger.info(
+                    f'Container {container.name} started, running initial setup...')
+                self._run_container_setup(container, target_adapter)
+                logger.info(f'Container {container.name} fully initialized.')
+
+                if len(arch_list) > 1:
+                    container.stop()
```
eculicny, line 188: Would it make sense to move this block of code back into something like the `get_stopped_container` function? It ends up being identical for both the `if` and the `else` blocks here.

snowshu/core/docker.py
```diff
@@ -144,20 +116,80 @@ def get_stopped_container(  # noqa pylint: disable=too-many-arguments
...
+                logger.info(
+                    f"Creating stopped container {tagged_container_name}...")
+                self.remove_container(tagged_container_name)
+
+                container = self.client.containers.create(
+                    image.tags[0],
+                    target_adapter.DOCKER_START_COMMAND,
+                    network=network.name,
+                    name=tagged_container_name,
+                    hostname=target_adapter._credentials.host,  # noqa pylint: disable=protected-access
```
eculicny, line 151: Can we not access `target_adapter.credentials` here? I thought that was wrapped with property decorators/functions.

eculicny (commented):
Looks good. I think we might be able to move some of the logic back into another function to reduce some identical code duplication.

snowshu/core/docker.py
skavierin: We can, it's just that underscored one caught my eye first

snowshu/core/docker.py
```diff
@@ -167,96 +199,75 @@ def get_stopped_container(  # noqa pylint: disable=too-many-arguments
...
+                    except AssertionError:
+                        logger.warning('Image tags do not match their actual architecture, '
+                                       'retag or delete postgres images manually to correct')
+
+                except ConnectionError as error:
+                    logger.error(
+                        'Looks like docker is not started, please start docker daemon\nError: %s', error)
+                    raise
+
+                tagged_container_name = f'snowshu_target_{arch}'
```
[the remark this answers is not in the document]
skavierin: We can't if we generalise this part as you asked below. This has to be target_adapter.credentials.hostname then

snowshu/core/docker.py
eculicny: Makes sense. Let's use the actual `credentials` attribute in this case then.

snowshu/core/docker.py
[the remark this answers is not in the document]
skavierin: Changed to `f'{hostname}_{arch}'`

snowshu/core/docker.py
skavierin: Moved into other method, together with setup.

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -185,7 +187,7 @@ def initialize_replica(self,
                         f"The override image is not a Postgres image: {incremental_image}")
             except Exception as error:
```
ArtemHU, line 188: should we define a custom exception for "image not available" case?
cc @eculicny

skavierin: That's a year old code, I think better not touch it now

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -41,9 +42,10 @@ def __init__(self, replica_metadata: dict):
 
         self.target = DOCKER_TARGET_CONTAINER if IS_IN_DOCKER else 'localhost'
         self.credentials = self._generate_credentials(self.target)
+        self.shdocker = SnowShuDocker()
```
ArtemHU, line 45: why don't to cast the SnowShuDocker class to static? cc @eculicny

eculicny (approved):
Looks good.

I think the updates suggested by @ArtemHU are probably worth looking into, but as separate pieces of work (out of scope of the hotfix)

PR #135, opened 2022-10-20
Reworked incremental difference to account for all connected nodes

eculicny:
Original logic was missing cases with longer dependency paths.


skavierin (approved):
LGTM

snowshu/core/graph.py
```diff
@@ -49,26 +48,17 @@ def catalog_difference(source_graph: Union["SnowShuGraph", networkx.Graph],
...
+        source_nodes = set(source_graph.nodes)
+        # relations not found in target catalog
+        missing_relations = source_nodes.difference(target_catalog)
+        # generate actual list since source_graph will be modified
+        weakly_connected_subgraphs = [comp for comp in networkx.weakly_connected_components(source_graph)]  # noqa pylint: disable=unnecessary-comprehension
 
-        isolated_relations_difference = set(isolated_relations).difference(target_catalog)
+        # if a component has no missing relations, remove those nodes from the graph
+        for component in weakly_connected_subgraphs:
+            if not missing_relations.intersection(component):
```
ArtemHU, line 59: btw, should we bump up the networkx? the actual one is 2.8.7 while we have 2.6.2

ArtemHU (approved):

snowshu/core/graph.py
eculicny: Probably worth updating most pins at some point. I think we only updated the pins recently for running snowshu on arm machines.

I'll create a ticket for doing that dependency update (along with dependency updates in other repos)

PR #141, opened 2022-12-27
DATA-3907: Implementation

skavierin:
- `copy_views_as_tables` added to replica file schema and template
- configuration_parser.py updated to pass this to source adapter
- test cases for for item 2

### DIFF
--- a/tests/unit/test_configuration_parser.py
+++ b/tests/unit/test_configuration_parser.py
@@ -2,13 +2,14 @@ import json
 import tempfile
 from io import StringIO
 from pathlib import Path
+import copy
 
 import pytest
 import yaml
 from jsonschema.exceptions import ValidationError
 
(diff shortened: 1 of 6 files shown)


tests/unit/test_configuration_parser.py
```diff
@@ -124,7 +130,44 @@ def test_schema_verification(tmpdir, stub_creds, stub_configs):
...
+        'undefined': {
+            'value': None,
+            'result': {
+                "BASE TABLE": materializations.TABLE,
+                "VIEW": materializations.TABLE
+            }
+        }
+    }
+
+    for case in cases.values():
```
ArtemHU, line 166: the https://docs.pytest.org/en/6.2.x/parametrize.html is more flexible&convenient approach to use 
every single iteration is treated a sep test case

ArtemHU (approved):

Gal40n04ek (approved):

PR #149, opened 2023-05-25
bugfix/DATA-4876/2: Change to pgdump + tests

skavierin:
Reworked copy process between replicas to use pgdump instead of brute force copy/paste of pgdata dir.
Cause of this change is Artem's finding that it is indeed impossible to kill postgres inside container because it is a main process of a container and if it exits container crashes.

Also renamed " docker_image" to "docker_image", dunno why there was a leading space, everything seems to work after.

Testing done:
- multiarch small replica build
- integration tests
- unit tests


ArtemHU:
> 

let's try to build a replica for dawarehouse/datalake following new approach

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -87,8 +86,10 @@ def __init__(self, replica_metadata: dict, **kwargs):
                                      f'-h {self._credentials.host} '
                                      f'-U {self._credentials.user} '
                                      f'-d {self._credentials.database}')
-        self.DOCKER_SHARE_REPLICA_DATA = f"cp -af $PGDATA/* {self.DOCKER_REPLICA_MOUNT_FOLDER}"  # noqa pylint: disable=invalid-name
-        self.DOCKER_IMPORT_REPLICA_DATA_FROM_SHARE = f"cp -R -f {self.DOCKER_REPLICA_MOUNT_FOLDER}/* $PGDATA/"  # noqa pylint: disable=invalid-name
+        self.DOCKER_SHARE_REPLICA_DATA = f"pg_dumpall -c -U {self._credentials.user} -p 9999 > " \
```
ArtemHU, line 89: should we achieve the output?

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -399,42 +400,28 @@ def statement_runner(statement: str):
...
             self.passive_container.start()
             logger.info('Copying replica data into passive container')
-            logger.info('Stopping postgres in passive container')
-            self.stop_postgres(self.passive_container)
 
-            logger.info("Purging passive container's pgdata dir")
-            self.passive_container.exec_run("/bin/bash -c 'rm -rf $PGDATA/*'", tty=True)
+            # Wait for db init
+            while True:
+                if self.passive_container.exec_run(self.DOCKER_READY_COMMAND).exit_code == 0:
```
ArtemHU, line 414: should we handle the case when this loop is running abnormally long ...
let's say docker is not ready due to some reason

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
skavierin: Don't understand the question. Achieve what?

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
skavierin: This is a textbook https://en.wikipedia.org/wiki/Halting_problem
How long should we wait for postgres spinup?

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
ArtemHU: the output of the command

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
ArtemHU: I believe ~10min is more than enough

tests/integration/postgres/test_postgres_target_adapter.py
```diff
@@ -156,10 +159,13 @@ def test_restore_data_from_shared_replica(docker_flush):
         envars=['POSTGRES_USER=snowshu',
                 'POSTGRES_PASSWORD=snowshu',
                 'POSTGRES_DB=snowshu',
-                f'PGDATA={DOCKER_REPLICA_MOUNT_FOLDER}'])
+                f'PGDATA=/pgdata'])
```
ArtemHU, line 162: should it be in uppercase?

skavierin: Does not matter, it's a dir name. Default ones under "/" are lower case

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -87,8 +86,11 @@ def __init__(self, replica_metadata: dict, **kwargs):
                                      f'-h {self._credentials.host} '
                                      f'-U {self._credentials.user} '
                                      f'-d {self._credentials.database}')
-        self.DOCKER_SHARE_REPLICA_DATA = f"cp -af $PGDATA/* {self.DOCKER_REPLICA_MOUNT_FOLDER}"  # noqa pylint: disable=invalid-name
-        self.DOCKER_IMPORT_REPLICA_DATA_FROM_SHARE = f"cp -R -f {self.DOCKER_REPLICA_MOUNT_FOLDER}/* $PGDATA/"  # noqa pylint: disable=invalid-name
+        self.DOCKER_SHARE_REPLICA_DATA = f"pg_dumpall -c -U {self._credentials.user} -p 9999 | " \
+                                         f"gzip > {self.DOCKER_REPLICA_MOUNT_FOLDER}/replica_dump.gz"  # noqa pylint: disable=invalid-name
```
ArtemHU, line 90: initially it was suggested to look into 
```
-F format
--format=format
Selects the format of the output. format can be one of the following:
```
https://www.postgresql.org/docs/12/app-pgdump.html

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -399,42 +401,31 @@ def statement_runner(statement: str):
...
 
-            logger.info("Purging passive container's pgdata dir")
-            self.passive_container.exec_run("/bin/bash -c 'rm -rf $PGDATA/*'", tty=True)
+            # Wait for db init
+            for _ in range(1200):  # for 10 minutes
+                if self.passive_container.exec_run(self.DOCKER_READY_COMMAND).exit_code == 0:
+                    break
+                time.sleep(.5)
+            else:
+                raise Exception(
```
ArtemHU, line 419: it is not a generic exception, I believe we should define a custom one
can be defined in the exceptions.py

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
skavierin: Probably if we dump databases individually, but that adds more complexity for an unclear benefit

tests/integration/postgres/test_postgres_target_adapter.py
ArtemHU: just a naming convention to follow

skavierin: DOCKER_REPLICA_MOUNT_FOLDER evaluates to "/app/snowshu_replica_data_shared", so it is still lower case. You might be confusing global variable name (all upper) to the path itself (generally lower). The change was made since pgdata content is no longer inside the volume (only the dump is).

snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
ArtemHU: I was asking about dumpall cmd

PR #151, opened 2023-11-10
snowflake create/drop commit for data-3496

KatzperWoz:
Hey, as this is my first ever encounter with snowshu I count on your help with onboarding me on that. I've created this initial commit so you can check what I've done for the ticket DATA-3496. I think there might be some changes to do, but rather drop it now so you can take a look and I could potentially improve it next week.

### DIFF
--- a/snowshu/adapters/source_adapters/snowflake_adapter.py
+++ b/snowshu/adapters/source_adapters/snowflake_adapter.py
@@ -129,8 +129,10 @@ class SnowflakeAdapter(BaseSourceAdapter):
                     {corrected_database}.{corrected_name}'''
             result = self._safe_query(query)
             logger.info("Schema creation result: %s", result['status'][0])
-        except Exception as err:
-            logger.error("An error occurred while creating the schema: %s", err)
+        except ValueError as err:
+            error_message = f"An error occurred while creating the schema {corrected_name} in database {corrected_database}: {err}"
+            logger.error(error_message)
+            return error_message
 
     def drop_schema(self, name: str, database: str = 'SNOWSHU') -> str:
         """Droop a schema and all of its contained objects (tables, views,
@@ -151,8 +153,10 @@ class SnowflakeAdapter(BaseSourceAdapter):
                     CASCADE'''
             result = self._safe_query(query)
             logger.info("Schema drop result: %s", result["status"][0])
-        except Exception as err:
-            logger.error("An error occurred while dropping the schema: %s", err)
+        except ValueError as err:
+            error_message = f"An error occurred while dropping the schema {corrected_name} in database {corrected_database}: {err}"
+            logger.error(error_message)
+            return error_message
 
     def create_table(self, query: str, name: str, schema: str, database: str = 'SNOWSHU'):
         corrected_name, corrected_schema, corrected_database = (
@@ -166,8 +170,10 @@ class SnowflakeAdapter(BaseSourceAdapter):
                          corrected_name, corrected_schema, corrected_database)
             result = self._safe_query(full_query)
             logger.info("Table creation result: %s", result['status'][0])
-        except Exception as err:
-            logger.error("An error occurred while creating the table: %s", err)
+        except ValueError as err:
+            error_message = f"An error occurred while creating the table {corrected_name} in database {corrected_database}: {err}"
+            logger.error(error_message)
+            return error_message
 
     def drop_table(self, name: str, schema: str, database: str = 'SNOWSHU'):
         corrected_name, corrected_schema, corrected_database = (
@@ -180,8 +186,10 @@ class SnowflakeAdapter(BaseSourceAdapter):
                          corrected_name, corrected_schema, corrected_database)
             result = self._safe_query(query)
             logger.info("Table drop result: %s", result["status"][0])
-        except Exception as err:
-            logger.error("An error occurred while dropping the table: %s", err)
+        except ValueError as err:
+            error_message = f"An error occurred while dropping the table {corrected_name} in database {corrected_database}: {err}"
+            logger.error(error_message)
+            return error_message
 
     @staticmethod
     def population_count_statement(relation: Relation) -> str:
--- a/snowshu/adapters/source_adapters/snowflake_adapter.py
+++ b/snowshu/adapters/source_adapters/snowflake_adapter.py
@@ -112,13 +112,13 @@ class SnowflakeAdapter(BaseSourceAdapter):
         logger.debug(f'Done. Found {len(tables)} tables in {schema} schema of {database} database.')
         return tables
 
-    def generate_schema(self, name: str, database: str = 'SNOWSHU_DEVELOPMENT') -> str:
+    def generate_schema(self, name: str, database: str = 'SNOWSHU') -> str:
         """Create a schema in the specified database.
 
             Args:
                 name: The name of the schema to create.
                 database: The database where the schema will be created.
-                          Defaults to 'SNOWSHU_DEVELOPMENT'.
+                          Defaults to 'SNOWSHU'.
         """
         corrected_database, corrected_name = (
             self._correct_case(x) for x in (database, name))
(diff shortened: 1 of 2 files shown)


snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -103,6 +103,100 @@ def _get_all_schemas(self, database: str, exclude_defaults: Optional[bool] = Fal
             f'Done. Found {len(schemas)} schemas in {database} database.')
         return schemas
 
+    def _get_all_tables(self, database: str, schema: str) -> List[str]:
+        database = self.quoted(database)
+        schema = self.quoted(schema)
+        logger.debug(f'Collecting tables from {schema} schema in {database} database in snowflake...')
+        show_result = self._safe_query(f'SHOW TERSE TABLES IN SCHEMA {database}.{schema}')['name'].tolist()
+        tables = list(set(show_result))
```
glysav, line 111: Are there any cases when SHOW TERSE TABLES IN SCHEMA return duplicate names?

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -103,6 +103,100 @@ def _get_all_schemas(self, database: str, exclude_defaults: Optional[bool] = Fal
...
+            query = f'''CREATE SCHEMA IF NOT EXISTS
+                    {corrected_database}.{corrected_name}'''
+            result = self._safe_query(query)
+            logger.info("Schema creation result: %s", result['status'][0])
+        except ValueError as err:
+            error_message = (
+                f"An error occurred while creating the schema {corrected_name} "
+                f"in database {corrected_database}: {err}"
+            )
+            logger.error(error_message)
```
glysav, line 137: Does it makes sense to continue execution(no re-raise here) if CREATE SCHEMA statement wasn't executed? 
Same concern for all functions introduced in this PR

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -103,6 +103,100 @@ def _get_all_schemas(self, database: str, exclude_defaults: Optional[bool] = Fal
...
+
+    def drop_schema(self, name: str, database: str = 'SNOWSHU'):
+        """Droop a schema and all of its contained objects (tables, views,
```
glysav, line 140: typo

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -103,6 +103,100 @@ def _get_all_schemas(self, database: str, exclude_defaults: Optional[bool] = Fal
...
+                f"An error occurred while dropping the schema {corrected_name} "
+                f"in database {corrected_database}: {err}"
+            )
+            logger.error(error_message)
+
+    def create_table(self, query: str, name: str, schema: str, database: str = 'SNOWSHU'):
+        corrected_name, corrected_schema, corrected_database = (
+            self._correct_case(x) for x in (name, schema, database)
+        )
+        full_query = f'''CREATE TABLE IF NOT EXISTS
```
ArtemHU, line 169: I suggest incorporating a table type into the query-template, and start with the transient table as the default option. 
Please refer to the sf doc for more info on this: https://docs.snowflake.com/en/sql-reference/sql/show-tables

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -103,6 +103,100 @@ def _get_all_schemas(self, database: str, exclude_defaults: Optional[bool] = Fal
...
+            result = self._safe_query(full_query)
+            logger.info("Table creation result: %s", result['status'][0])
+        except ValueError as err:
+            error_message = (
+                f"An error occurred while creating the table {corrected_name} "
+                f"in database {corrected_database}: {err}")
+            logger.error(error_message)
+
+    def drop_table(self, name: str, schema: str, database: str = 'SNOWSHU'):
+        corrected_name, corrected_schema, corrected_database = (
```
ArtemHU, line 184: I am curios about the specific use case for this function.

tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -324,3 +324,72 @@ def test_check_count_and_query(sf_adapter):
 
     assert exc.errisinstance(TooManyRecords)
     assert sf_adapter.check_count_and_query.retry.statistics["attempt_number"] == 4
+
+
+def test_generate_schema(sf_adapter):
+    DATABASE, SCHEMA = "SNOWSHU_DEVELOPMENT", "GENERATE_SCHEMA_TEST"
```
ArtemHU, line 330: It' good to provide a short description for every single test case we are adding

tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -324,3 +324,72 @@ def test_check_count_and_query(sf_adapter):
...
+        sf_adapter.drop_schema(database=DATABASE, name=SCHEMA)
+
+
+def test_drop_schema(sf_adapter):
+    DATABASE, SCHEMA = "SNOWSHU_DEVELOPMENT", "DROP_SCHEMA_TEST"
+    try:
+        # Setup: Ensure the schema exists before attempting to drop it
+        if SCHEMA not in sf_adapter._get_all_schemas(DATABASE):
+            sf_adapter.generate_schema(database=DATABASE, name=SCHEMA)
+
```
ArtemHU, line 350: one more assert to ensure that schema exists, before running the actual drop statement

tests/integration/snowflake/test_snowflake_adapter.py
```diff
@@ -324,3 +324,72 @@ def test_check_count_and_query(sf_adapter):
...
+        # Test table dropping
+        sf_adapter.drop_table(name=TABLE, schema=SCHEMA, database=DATABASE)
+        assert TABLE not in sf_adapter._get_all_tables(DATABASE, SCHEMA)
+    finally:
+        # Clean up: Drop table and schema if they still exist
+        if TABLE in sf_adapter._get_all_tables(DATABASE, SCHEMA):
+            sf_adapter.drop_table(name=TABLE, schema=SCHEMA, database=DATABASE)
+        if SCHEMA in sf_adapter._get_all_schemas(DATABASE):
+            sf_adapter.drop_schema(name=SCHEMA, database=DATABASE)
```
ArtemHU, line 395: I believe it's enough to run the drop schema statement since it has the cascade option

snowshu/adapters/source_adapters/snowflake_adapter.py
KatzperWoz: Normally I'd say not, as Snowflake forces name uniqueness, though I found one bypass, it's using quotes to create tables eg:

```
CREATE TABLE "myTable" (id INT);
CREATE TABLE "MYTABLE" (id INT);  -- This will succeed
```

Do you see a scenario where it can com back to bite us?

snowshu/adapters/source_adapters/snowflake_adapter.py
KatzperWoz: As I'm still deep diving and exploring the entire idea of snowshu project and how it all works together I think I might not know exactly correct answer for that. Nevertheless I was thinking about it also, why we might need it, if we have `drop_schema w/cascade`, but I decided that for now I'm gonna follow ticket specs, cause I assumed that there might be a possibility that in the future if we notice some discrepancy between the state of a current replica and current config, we will want to drop just the removed table and (if specified) add a new one, without dropping whole schema and rebuilding from scratch. What do you think about it? Does that make sense?

snowshu/adapters/source_adapters/snowflake_adapter.py
KatzperWoz: Yeah, good observation I might have missed that..

tests/integration/snowflake/test_snowflake_adapter.py
KatzperWoz: That's right, unnecessary duplication here

ArtemHU (approved):

glysav (approved):

PR #152, opened 2023-12-05
[Small] Feat/data 3496 add create drop functionality

KatzperWoz:
Create a function in the utils library that adds a unique ID to a name. This prevents the same name being used more than once, especially when creating snowshu replicas in snowflake.

More details regarding ticket can be found here: https://health-union.atlassian.net/browse/DATA-3497

### DIFF
--- a/snowshu/core/utils.py
+++ b/snowshu/core/utils.py
@@ -1,5 +1,6 @@
 import os
 import re
+import uuid
 from importlib import import_module
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional, TextIO, Type, Union, List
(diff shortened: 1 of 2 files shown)


snowshu/core/utils.py
```diff
@@ -122,3 +123,9 @@ def remove_dangling_replica_containers() -> None:
     for container in client.containers.list(all=True):
         if 'snowshu_target_' in container.name:
             container.remove(force=True)
+
+
+def generate_unique_uuid_name(name: str) -> str:
+    """Generates a unique name based on name and randomly generated uuid."""
+    _uuid = str(uuid.uuid4()).rsplit('-', maxsplit=1)[-1]
```
KatzperWoz, line 130: I'm using only the last part of the UUID for two main reasons: a full UUID is too long and looks awkward, and using just a part of it makes it easier to identify your replicas while still keeping the chance of name conflicts very low.

glysav (approved):

ArtemHU (approved):

PR #154, opened 2023-12-08
Feat/data 3498 switch sampling queries

KatzperWoz:
Tl;dr: Switch sampling queries, instead of querying sampled data we first create a transient table with sampled data and then fetch transient table to local memory. On top of that some code refactors.

Context:
The goal is to transition from storing temporary data in memory to storing it in Snowflake. 

During data extraction, Snowshu should utilize a previously created function to generate schemas and tables in Snowflake. After creating these schemas and tables, it should transfer data from the temporary tables into the relation.data attribute.

After processing the entire connected subgraph, the temporary schemas and tables in Snowflake should be cleaned up. 

The core part responsible for it is placed in ```graph_set_runner.py``` module.

On top of that there are some refactors making code easier to read and understand.

Test Plan:
If someone would like to test it, I'd recommend to install this package from source and later try to execute snowshu create function on one of snowflake repos eg. hu-dw-artifact, hu-dw-datawarehouse

### DIFF
--- a/snowshu/adapters/source_adapters/snowflake_adapter.py
+++ b/snowshu/adapters/source_adapters/snowflake_adapter.py
@@ -532,13 +532,11 @@ LIMIT {max_number_of_outliers})
                 'Adapter.get_connection called before setting Adapter.credentials')
 
         logger.debug(f'Acquiring {self.CLASSNAME} connection...')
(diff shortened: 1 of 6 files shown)


Gal40n04ek (approved):

glysav (approved):
This changes looks good to me. Works fine locally!

PR #155, opened 2023-12-13
Feat/data 3499 switch sampling references

KatzperWoz:
The main goal of this PR is to swich compilation of predicate_constraint_statement in snowflake instead of relation.data. On top of that it required a series of refactors to make it possible.

This Change:

- Refactor of predicate_constraint_statement and polymorphic_constraint_statement to make it work on Snowflake, not locally
- Refactor graph_set_runner and Relation to contain temp_schema name
- Refactor test_compiler.py so it doesn't rely on downstream methods
- Refactor test_snowflake_adapter.py to adjust to changes

Test Plan:
To test it, I'd recommend to run creation of snowshu replica in hu-dw-warehouse, including one general_relation and 3 different specified_relations (directional, bidirectional, polymorphic). Regarding massive amount of refactor changes in test_compiler.py, I'd recommend to check just 2-3 tests - the changes are similar in every test so it's more to understand whether suggested approach is acceptable.

### DIFF
--- a/snowshu/adapters/source_adapters/snowflake_adapter.py
+++ b/snowshu/adapters/source_adapters/snowflake_adapter.py
@@ -388,7 +388,7 @@ LIMIT {max_number_of_outliers})
                                          remote_key: str,
                                          local_type: str,
                                          local_type_match_val: str = None) -> str:
-        predicate = self.predicate_constraint_statement(relation, analyze, local_key, remote_key)
+        predicate = SnowflakeAdapter().predicate_constraint_statement(relation, analyze, local_key, remote_key)
         if local_type_match_val:
             type_match_val = local_type_match_val
         else:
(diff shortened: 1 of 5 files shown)


Gal40n04ek (approved):

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -335,49 +335,61 @@ def upstream_constraint_statement(relation: Relation,
...
+                )
+                if constraint_set.empty:
+                    raise ValueError(
+                        f"The constraint set for remote key {remote_key} "
+                        f"in {relation.temp_dot_notation} is empty."
+                    )
+                constraint_set = constraint_set.iloc[:, 0].tolist()[
+                    : SnowflakeAdapter.SNOWFLAKE_MAX_NUMBER_EXPR
+                ]
+                constraint_sql = ",".join(f"'{item}'" for item in constraint_set)
```
ArtemHU, line 363: why do we need to go with the trouble of building the list of keys like this(via a huge string)? can't we just switch it to a subquery? After all, all our processing steps are supposed to go to the snowflake.

ArtemHU (approved):

PR #156, opened 2023-12-21
Switch to local data frames instead of relation.data

KatzperWoz:
This pull request refactors the load_data_into_relation method to handle local dataframes. It also adjusts the functions and tests that use this method accordingly.

### DIFF
--- a/snowshu/adapters/target_adapters/base_target_adapter.py
+++ b/snowshu/adapters/target_adapters/base_target_adapter.py
@@ -106,22 +106,32 @@ class BaseTargetAdapter(BaseSQLAdapter):  # noqa pylint: disable=too-many-instan
         raise NotImplementedError()
 
     def load_data_into_relation(self, relation: Relation, data: pd.DataFrame) -> None:
-        """Loads data into a to target.
+        """Loads data into a target.
 
         Args:
             relation: The relation containing info about dataset to load.
             data: The data to load into the relation.
         """
-        if data is None and relation.data.empty:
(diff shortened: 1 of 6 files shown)


snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -103,33 +105,55 @@ def create_or_replace_view(self, relation) -> None:
...
-    def load_data_into_relation(self, relation: Relation) -> None:
+    def load_data_into_relation(self, relation: Relation, data: pd.DataFrame) -> None:
+        """Loads data into a to target.
+
+        Args:
+            relation: The relation containing info about dataset to load.
+            data: The data to load into the relation.
+        """
+        if data is None and relation.data.empty:
+            raise ValueError("Both data and relation.data are empty.")
```
glysav, line 116: Why the case when relation have no data is invalid? For sure it doesn't seem very useful, but it still can be valid, am I wrong?

glysav (commented):
One question regarding no data in relation, but overall it looks good to me

snowshu/adapters/target_adapters/base_target_adapter.py
KatzperWoz: You're right, I've fixed it to not raise exception but also added some logs to be more informative on replica creation if happens with empty data.

glysav (approved):

Gal40n04ek (approved):

PR #157, opened 2024-01-22
Hotfix/snowflake adapter missing credentials

KatzperWoz:
It's a hotfix to issue with _polymorphic_constraint_statement_, related to this: [https://us-east-1.console.aws.amazon.com/codesuite/codebuild/251286366697/projects/dev-snowsh[…]45e616-c83f-4800-8357-fe4144e838c0/?region=us-east-1](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/251286366697/projects/dev-snowshu-replica-builder-build-develop/build/dev-snowshu-replica-builder-build-develop%3A4745e616-c83f-4800-8357-fe4144e838c0/?region=us-east-1). It's in warehose_dev account, where in logs one need to look for '_credentials'.

### DIFF
--- a/snowshu/adapters/source_adapters/snowflake_adapter.py
+++ b/snowshu/adapters/source_adapters/snowflake_adapter.py
@@ -348,6 +348,7 @@ LIMIT {max_number_of_outliers})
                 f"FROM ("
                 f"    SELECT DISTINCT {remote_key} "
                 f"    FROM {relation.temp_dot_notation} "
+                f"    WHERE {remote_key} NOT LIKE '%''%' "
                 f"    LIMIT {SnowflakeAdapter.SNOWFLAKE_MAX_NUMBER_EXPR}"
                 f") AS subquery"
             )
--- a/snowshu/adapters/source_adapters/snowflake_adapter.py
+++ b/snowshu/adapters/source_adapters/snowflake_adapter.py
@@ -344,7 +344,7 @@ LIMIT {max_number_of_outliers})
                 return f"{local_key} IN ( SELECT {remote_key} AS {local_key} FROM ({relation.core_query}))"
 
             constraint_query = (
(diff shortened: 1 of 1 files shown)


snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -348,6 +348,7 @@ def predicate_constraint_statement(
                 f"FROM ("
                 f"    SELECT DISTINCT {remote_key} "
                 f"    FROM {relation.temp_dot_notation} "
+                f"    WHERE {remote_key} NOT LIKE '%''%' "
```
KatzperWoz, line 351: While testing fixed solution, I encountered an edge case where value in returned list contained single quote eg. **kacper.w'ozniak@mail.com** - this was causing failing sample table creation, hence I added a heuristic to avoid values containing single quote. (It's hard to replicate the issue as every time it's sampling a different subset)

Gal40n04ek (approved):

snowshu/adapters/source_adapters/snowflake_adapter.py
ArtemHU: Why don't we try escaping quotes in the case of string values, instead of excluding them?

KatzperWoz: I was thinking that the amount of such edge cases is minor and if we escape it and then later try to use it again to filter, it may not match the pattern or cause some other downstream issues, hence excluding this minor number of records containing quotes seemed to be simpler and safer in my opinion.

ArtemHU: Since we are implementing this on the Snowflake side, it might make more sense to pass a query instead of a list of keys, what do you think?

PR #159, opened 2024-03-07
DATA-6876: fix casting for predicates, add/modify tests

KatzperWoz:


eculicny:
Would it make sense to have the cast be to the type of the local attribute instead of always to a `varchar`?

Also, why reverse the booleans for the data types? I think I'm missing why that's necessary (if using it), and it's a bit confusing if reversed and used that way. Wouldn't we want the `local_attribute`'s type applied to the `remote_attribute`'s type here?

KatzperWoz:
> Would it make sense to have the cast be to the type of the local attribute instead of always to a `varchar`?
> 
> Also, why reverse the booleans for the data types? I think I'm missing why that's necessary (if using it), and it's a bit confusing if reversed and used that way. Wouldn't we want the `local_attribute`'s type applied to the `remote_attribute`'s type here?

@eculicny I think I took to big shortcut with this `varchar`, though changed it now so `remote_key` is always casted to `local_key` type. 

The logic behind reversed booleans was that,  now we don't need to put `quoted` values into quotes (which would result in snowflake errors), but we actually convert `unquoted` values to `varchar` which is kinda a quoted/string value - nevertheless, it doesn't make sense anymore with current implementation, so I reverted original approach.

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -367,16 +367,47 @@ def _validate_key_index_error(self,
                 f"Remote key {remote_key} not found in {relation.temp_dot_notation} table."
             ) from err
 
+    def format_remote_key(
+        self, relation: Relation, local_key: str, remote_key: str
+    ) -> str:
+        """Formats the remote key based on whether it needs to be quoted or not."""
+
+        local_key_type_query = (
```
glysav, line 375: It seems like we can put 375-387 lines inside the `if` statement. So the SQL which find `local_key_type` will be executed only in the case when we need it, instead of always.

snowshu/adapters/source_adapters/snowflake_adapter.py
```diff
@@ -367,16 +367,45 @@ def _validate_key_index_error(self,
                 f"Remote key {remote_key} not found in {relation.temp_dot_notation} table."
             ) from err
 
+    def format_remote_key(
+        self, relation: Relation, local_key: str, remote_key: str
+    ) -> str:
+        """Formats the remote key based on whether it needs to be quoted or not."""
+        attribute = relation.lookup_attribute(remote_key)
+        if not attribute.data_type.requires_quotes:
```
eculicny, line 375: If `attribute` is not `None`, shouldn't the `data_type` value already be populated? Then we wouldn't need to call out to Snowflake for the type.

If we still need to make the query to Snowflake, we should probably add `not attribute` to the if condition to avoid an attribute error.

KatzperWoz: > If attribute is not None, shouldn't the data_type value already be populated? Then we wouldn't need to call out to Snowflake for the type.

Actually I think that neither yours suggestion nor my current solution is gonna work. After running some extra tests seems like it is impossible to retrieve `local_key_type` from snowflake as our `Relation` object doesn't contain information on database and table name where our column is located and as we need to cast remote key to local key type, there might be a case where `lookup_attribute(local_key)` is `None` and we just simply cannot get it - this is the case reported by Luciano.

> If we still need to make the query to Snowflake, we should probably add not attribute to the if condition to avoid an attribute error.

This one is irrelevant for us now.

> Edit: Would it also make sense to short circuit the logic here to only cast when the types don't match as opposed to the requires_quotes flag since string:string comparison doesn't need a type cast?

I think this one is not super easy thing to do, as we cannot just simply compare `lookup_attribute(remote_key)` to `lookup_attribute(local_key)` as we might get None for one of these. We'd need to cache somewhere `lookup_attribute(local_key)` type before running predicates to use it later. Or am I missing some simple solution here?

Ps. for now the only solution that was simple and was working (though) was not elegent is still one below:
```
    def format_remote_key(
        self, relation: Relation, local_key: str, remote_key: str
    ) -> str:
        """Formats the remote key based on whether it needs to be quoted or not."""
        attribute = relation.lookup_attribute(remote_key)
        if not attribute.data_type.requires_quotes:
               return f"{remote_key}::VARCHAR"
        return remote_key
```

KatzperWoz: Ps2. I don't make any changes for now, before we agree on one solution as adjusting tests is too time consuming to make changes all the time :)

eculicny: No worries. In the past, I know there were uses of the `Relation` model that didn't actually have all of the details for the relations, so I'm not surprised we're running into it here since that data hadn't been needed previously.

I would say, let's go back to your original hotfix since it worked and likely solves the primary type mismatch case of numeric-string. I'd leave the `requires_quotes` boolean as is instead of flipping it, but maybe making the description more detailed could help avoid confusion.

We can spin off some other work to investigate more since we probably would need to adjust how the graphs/catalogs are built if we're missing fully-qualified relation info.

glysav (approved):
Since it was decided to roll back to the fix of the primary type mismatch case -- this looks good to me

KatzperWoz:
@eculicny any further concerns, or we can push it?

eculicny (approved):
No other concerns given the current plan.

PR #160, opened 2024-03-13
Bugfix/data 6675 failed to load relation to target

KatzperWoz:


snowshu/core/graph_set_runner.py
```diff
@@ -173,8 +174,9 @@ def _traverse_and_execute(self, executable: GraphExecutable) -> None:  # noqa: p
                 f"Executing graph with {len(executable.graph)} relations in it...")
             sorted_graphs = nx.algorithms.dag.topological_sort(executable.graph)
             for i, relation in enumerate(sorted_graphs, start=1):
-
-                relation.temp_schema = "_".join([relation.schema, self.uuid])
+                relation.temp_schema = "_".join(
```
KatzperWoz, line 177: Here's major change, we needed to make schemas unique, due to the fact that we encountered a scenario where we process realtions that comes from different databases, but the same schemas and table names causing naming collisions.

snowshu/core/graph_set_runner.py
```diff
@@ -222,26 +224,49 @@ def _traverse_and_execute(self, executable: GraphExecutable) -> None:  # noqa: p
...
+                                    relation.unsampled,
+                                )
+                            )
+                            relation.sample_size = len(query_data)
+                            logger.info(
+                                f"{relation.sample_size} records retrieved for relation {relation.dot_notation}."
+                            )
+                        # This except block is necessary due to VARIANT data type issues
+                        # in Snowflake. In the future, we should remove this and find a
+                        # better solution.
```
KatzperWoz, line 252: Temporary exception handling so the process doesn't fail but gives us information on where the issue happened. Issue is going to be investigated in separate ticket.

PR #161, opened 2024-03-13
Bugfix/data 6675 failed to load relation to target

KatzperWoz:


snowshu/core/graph_set_runner.py
```diff
@@ -173,8 +174,9 @@ def _traverse_and_execute(self, executable: GraphExecutable) -> None:  # noqa: p
                 f"Executing graph with {len(executable.graph)} relations in it...")
             sorted_graphs = nx.algorithms.dag.topological_sort(executable.graph)
             for i, relation in enumerate(sorted_graphs, start=1):
```
KatzperWoz, line 176: Here's major change, we needed to make schemas unique, due to the fact that we encountered a scenario where we process realtions that comes from different databases, but the same schemas and table names causing naming collisions.

snowshu/core/graph_set_runner.py
```diff
@@ -173,8 +174,9 @@ def _traverse_and_execute(self, executable: GraphExecutable) -> None:  # noqa: p
...
-
-                relation.temp_schema = "_".join([relation.schema, self.uuid])
+                relation.temp_schema = "_".join(
+                    [relation.database, relation.schema, self.uuid]
+                )
```
KatzperWoz, line 179: Here's major change, we needed to make schemas unique, due to the fact that we encountered a scenario where we process realtions that comes from different databases, but the same schemas and table names causing naming collisions.

snowshu/core/printable_result.py
```diff
@@ -29,41 +29,63 @@ def to_tuple(self) -> list:
...
+                or (relation.population_size < relation.sampling.size)
+                else relation.sampling.size
+            )
+            percent = int(round(100.0 * (relation.sample_size / target_sample_size)))
+
+        percent_is_acceptable = (
+            True if isinstance(percent, str) else abs(percent - 100) <= 5
+        )
+        sample_size = relation.sample_size
+    except AttributeError as exc:
```
KatzperWoz, line 61: Another temporary exception handling, due to JSONDecodeError, relation is missing some attributes causing this function to fail, we handle it so the process doesn't fail, but gives us information in the final table about missing tables.

Gal40n04ek (approved):

PR #162, opened 2024-03-15
Bugfix/data 6671 failed bidirectional relationship

KatzperWoz:
Below relation is causing a following issue: `ValueError: failure in building row for relation DATA_LAKE.GOOGLE_ADS_CURRENT.GOOGLE_ADS_CURRENT_GEO_PERFORMANCE_REPORT : 'Relation' object has no attribute 'sample_size'` 

```
    - database: DATA_LAKE
      schema: GOOGLE_ADS_CURRENT
      relation: "(?i)^GOOGLE_ADS_CURRENT_(AD_PERFORMANCE_REPORT|AGE_RANGE_PERFORMANCE_REPORT|GEO_PERFORMANCE_REPORT|GENDER_PERFORMANCE_REPORT|KEYWORDS_PERFORMANCE_REPORT|SEARCH_QUERY_PERFORMANCE_REPORT|USER_LOCATION_PERFORMANCE_REPORT)$"
      relationships:
        bidirectional:
          - local_attribute: AD_GROUP_ID
            database: DATA_LAKE
            schema: GOOGLE_ADS_CURRENT
            relation: GOOGLE_ADS_CURRENT_AD_GROUPS
            remote_attribute: ID
```

From my research it seems like the reason is setting ```do_not_sample``` for ```bidirectional``` relations, I think that initially we assumed that building a predicate eg. ```ID IN (SELECT DISTINCT CONTACT_ID FROM SAMPLED_TABLE)``` is going to be enough to narrow down the upstream dataset. Meanwhile the relation above has a table where this assumption is not fulfilled and the predicate is not enough to narrow down/sample upstream table, resulting in error. I offered the simplest solution with removing the assumption and running sampling each time, though I think if we'd want to stand by the current approach we can add some exception handling where we run this sampling for bidirectional only if we catch the exception of TooManyRecords. I'm dropping this quick solution now to discuss and will be working/testing in the meantime on the other solution. 

Ps. I'll adjust tests once we agree on the solution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,5 +10,6 @@ apt-get install -y --reinstall build-essential && \
 apt-get install -y gcc && \
 apt-get install -y vim nano htop screen && \
 echo "export PYTHONPATH=${PWD}:$PYTHONPATH" >> /root/.bashrc && \
-pip3 install -r requirements/dev.txt && \
-pip3 install .
+pip3 install uv \
+uv pip install -r requirements/dev.txt && \
+uv pip install -e .
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,5 +11,5 @@ apt-get install -y gcc && \
 apt-get install -y vim nano htop screen && \
 echo "export PYTHONPATH=${PWD}:$PYTHONPATH" >> /root/.bashrc && \
 pip3 install uv \
-uv pip install -r requirements/dev.txt && \
-uv pip install -e .
+uv pip install --system -r requirements/dev.txt && \
+uv pip install --system -e .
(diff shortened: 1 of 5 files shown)


ArtemHU (approved):

KatzperWoz:
@eculicny @ArtemHU I'm sticking to initial solution (removal 'bidirectional' from if statement), as the other solution seemed to be no improvement, it implemented just extra lines of code, complexity and no efficiency improvement, hence I'm staying with this one. I adjusted test suite accordingly and as a bonus work I've switched package install manager from ```pip``` to ```uv``` that gives us nice improvement of ~30-40s on package installation process.

eculicny (approved):

PR #164, opened 2024-04-15
Feat/split responsibilities base target adapter class

KatzperWoz:
[DATA-7085](https://health-union.atlassian.net/browse/DATA-7085) and [DATA-7086](https://health-union.atlassian.net/browse/DATA-7086) are the tickets associated with the pull request. The primary objective of this pull request was to split BaseTargetAdapter into two more specific subclasses: BaseLocalTargetAdapter and BaseRemoteTargetAdapter. This segregation aims to establish an adaptable interface, allowing us to build atop adapters meant for local execution (e.g., Postgres) or others designed for use with external systems (e.g., Snowflake).

I am currently submitting the PR to the feature branch feat/develop-snowflake-target-adapter in order to prevent unnecessary or obsolete source code from being stored on the main branch.

Looking forward to your feedback.

[DATA-7085]: https://health-union.atlassian.net/browse/DATA-7085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DATA-7086]: https://health-union.atlassian.net/browse/DATA-7086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -152,156 +105,90 @@ def load_data_into_relation(self, relation: Relation, data: pd.DataFrame) -> Non
...
-            self, dialect: str, database: Optional[str] = None) -> str:
-        database = database if database is not None else self._credentials.database
-        conn_string = (f"{self.dialect}://{self._credentials.user}:{self._credentials.password}"
-                       f"@{self._credentials.host}:{self.DOCKER_TARGET_PORT}/{database}?")
-        return conn_string, {USER, PASSWORD, HOST, PORT, DATABASE, }
-
-    def _get_data_type(self, source_type: str) -> DataType:
         try:
-            return self.DATA_TYPE_MAPPINGS[source_type.replace(' ', '_').lower()]
+            return self.DATA_TYPE_MAPPINGS.get(source_type.replace(" ", "_").lower())
```
eculicny, line 131: Switching `.get` here will end up returning `None` instead of erroring out as intended.

snowshu/core/graph_set_runner.py
```diff
@@ -27,7 +28,7 @@
 class GraphExecutable:
     graph: nx.Graph
     source_adapter: BaseSourceAdapter
-    target_adapter: BaseTargetAdapter
+    target_adapter: Union[BaseLocalTargetAdapter, BaseRemoteTargetAdapter]
```
eculicny, line 31: Did this one actually need to change?

snowshu/adapters/base_sql_adapter.py
```diff
@@ -19,7 +20,7 @@
 logger = logging.getLogger(__name__)
 
 
-class BaseSQLAdapter:
+class BaseSQLAdapter(ABC):
```
ArtemHU, line 23: why don't to mark all abstracted methods with the @abc.abstractmethod decorator, document the expected behaviour and eliminate the explicit `NotImplementedError` exception?
https://docs.python.org/3/library/abc.html#abc.abstractmethod

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -152,156 +105,90 @@ def load_data_into_relation(self, relation: Relation, data: pd.DataFrame) -> Non
...
         except KeyError as err:
             logger.error(
-                '%s adapter does not support data type %s.', self.CLASSNAME, source_type)
+                f"{self.CLASSNAME} adapter does not support data type {source_type}."
```
ArtemHU, line 134: better to revert the change, for more details please refer to the https://medium.com/swlh/why-it-matters-how-you-log-in-python-1a1085851205

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -152,156 +105,90 @@ def load_data_into_relation(self, relation: Relation, data: pd.DataFrame) -> Non
...
+                self.CLASSNAME,
+            )
+        except Exception as exc:
+            logger.error(
+                "Error applying function %s to target %s: %s",
+                function,
+                self.CLASSNAME,
+                exc,
+            )
+            raise
```
ArtemHU, line 194: should we change it `raise from` to provide more context about why the exception occurred?

snowshu/core/graph.py
```diff
@@ -421,31 +421,52 @@ def build_sum_patterns_from_configs(config: Configuration) -> List[dict]:
...
+                    if lower_level.database_pattern
+                    else upper_level.database_pattern,
+                    schema=lower_level.schema_pattern
+                    if lower_level.schema_pattern
+                    else upper_level.schema_pattern,
+                    name=lower_level.relation_pattern,
+                )
+                approved_second_level_specified_patterns.append(pattern)
+
+        all_patterns = (
```
ArtemHU, line 464: I believe you have applied formatting to the code. I am curious, what was the reason?

snowshu/adapters/target_adapters/base_target_adapter.py
KacpiW: I agree

snowshu/core/graph.py
KacpiW: I refactored it to regular loops instead of deeply nested list comprehensions for the sake of readability and ease of understanding. I think previous version substantially sacrificed readability for the sake of slight performance gain, which in my opinion was a bit overengineered, but I'm open for discussion.

snowshu/adapters/base_sql_adapter.py
KacpiW: We can actually do that, fair point. We just need to be aware, that we'll have to be explicit and apply these methods in each or add NotImplementedError() if not applicable to every child class. I'd offer to merge this PR without this refactor, I'll make a separate one after this one is approved to avoid growing this PR more and more, is that ok for you?

snowshu/core/graph_set_runner.py
KacpiW: good point, it doesn't I think I wanted to be too specific

snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -152,156 +105,90 @@ def load_data_into_relation(self, relation: Relation, data: pd.DataFrame) -> Non
...
+            logger.error(
+                "Exception encountered loading data into %s: %s",
+                self.quoted_dot_notation(relation),
+                exc,
+            )
             raise
 
         logger.info(final_message)
 
-    def initialize_replica(self,
```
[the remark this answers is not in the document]
KatzperWoz: Yeah, that is good point to have in mind. I think I still have a mindset that `replica == docker`, thus I pushed it down the road to `LocalAdapter` subclass which might not be the right way. I'm not gonna make any changes for now with it, but I added this comment to future [ticket](https://health-union.atlassian.net/browse/DATA-7088) related to mentioned refactor to keep that in mind and figure it out. Though thanks for that idea ))

snowshu/adapters/base_sql_adapter.py
ArtemHU: As far as I am aware, the explicit invocation of NotImplementedError() is an outdated approach. I have noticed that you have partially transitioned to a different approach, which is why I have raised this concern in order to maintain consistency within the project.

ArtemHU (approved):

eculicny (approved):
Looks fine to me. Let's just make sure the few changes we've set for the following ticket/PR get noted and added.

PR #166, opened 2024-04-17
Refactor/align methods to abc design

KatzperWoz:
Aligning BaseSQLAdapter to ABC.

### DIFF
--- a/tests/unit/test_base_adapter.py
+++ b/tests/unit/test_base_adapter.py
@@ -10,94 +10,93 @@ from snowshu.core.models.credentials import (ACCOUNT, DATABASE, HOST, PASSWORD,
 from tests.common import rand_string
 
 
+REQUIRED_CREDENTIALS = (USER, PASSWORD, HOST)
+ALLOWED_CREDENTIALS = (ACCOUNT, SCHEMA)
+
+class StubbedAdapter(BaseSQLAdapter):
+    REQUIRED_CREDENTIALS = REQUIRED_CREDENTIALS
+    ALLOWED_CREDENTIALS = ALLOWED_CREDENTIALS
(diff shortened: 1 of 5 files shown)


tests/unit/test_base_adapter.py
```diff
@@ -10,94 +10,93 @@
...
+    )
+])
+def test_build_catalog(stubbed_adapter, config_patterns, mock_filtered_schema, included_relations, excluded_relations):
     mock_relations = included_relations + excluded_relations
-    def mock_get_relations_func(schema_obj: BaseSQLAdapter._DatabaseObject):
+    def mock_get_relations_func(schema_obj: StubbedAdapter._DatabaseObject):
         return [r for r in mock_relations if schema_obj.full_relation.schema == r.schema]
 
-    # stubbed version of the BaseSourceAdapter with the required class vars
-    class StubbedSourceAdapter(BaseSourceAdapter):
```
KatzperWoz, line 88: @eculicny do you know what's the exact reason of having this class here? I refactored class a little, seems to be working as expected but for now removed it as didn't see a point, but maybe I'm missing something?

ArtemHU (approved):

tests/unit/test_base_adapter.py
eculicny: Currently, I'm not sure what the original reason was. Probably either setup for future tests or tests that no longer exist.

I haven't checked, but I assume it isn't referenced anywhere else.

eculicny (approved):

tests/unit/test_base_adapter.py
KatzperWoz: No, this is the only place it exists. I see no reference to that

PR #167, opened 2024-04-22
Feat/add snowflake target adapter class

KatzperWoz:


snowshu/adapters/snowflake_common.py
```diff
@@ -0,0 +1,57 @@
...
+
+        By default, uses the instance credentials unless database or
+        schema override are provided.
+        """
+        if not self._credentials:
+            raise KeyError(
+                "Adapter.get_connection called before setting Adapter.credentials"
+            )
+
+        logger.debug(f"Acquiring {self.CLASSNAME} connection...")
```
glysav, line 29: Is the `self.CLASSNAME` existing attribute? Shouldn't it be something like `self.__class__.__name__` ?

glysav: I see it was inherited from BaseSqlAdapter [here](https://github.com/Health-Union/snowshu/blob/91a08038beef2298193d02195f7fdd1473c211ba/snowshu/adapters/base_sql_adapter.py#L43) earlier, but this class doesn't inherit it, so looks like this should be changed

snowshu/adapters/target_adapters/base_local_target_adapter.py
```diff
@@ -160,9 +160,7 @@ def create_all_database_extensions(self):
         """Create all required database extensions for the target adapter."""
 
     @abstractmethod
-    def initialize_replica(
-        self, source_adapter_name: str, incremental_image: str = None
-    ) -> None:
+    def initialize_replica(self, source_adapter_name: str, **kwargs) -> None:
```
glysav, line 163: If we put `incremental_image` into kwargs, let's change the docstring accordingly

tests/unit/test_replica_factory.py
```diff
@@ -3,13 +3,16 @@
 from io import StringIO
 from pathlib import Path
 from unittest import mock
+from webbrowser import get
```
glysav, line 6: What is this? :)

snowshu/adapters/snowflake_common.py
KatzperWoz: Good catch, I copied it but didn't notice that

snowshu/adapters/target_adapters/base_remote_target_adapter.py
```diff
@@ -1,5 +1,24 @@
+import logging
+from io import StringIO
+from typing import Union
+
+from snowshu.core.configuration_parser import ConfigurationParser, TEMPLATES_PATH
```
eculicny, line 5: Do we need to reference the config parser here? Would it make more sense to have the replica factory pass in the credentials after they are run through the config parser?

snowshu/core/configuration_parser.py
```diff
@@ -108,10 +108,10 @@ class ConfigurationParser:
     default_case: str = None
     preserve_case: bool = False
 
-    def _get_dict_from_anything(self,
-                                dict_like_object: Union[str, 'StringIO', dict],
-                                schema_path: Path) -> dict:
-        """Returns dict from path, io object or dict.  
+    def get_dict_from_anything(self,
```
eculicny, line 111: Is this only moving from public to private due to its use in the new adapter?

snowshu/templates/credentials_schema_target.json
eculicny, line 1: Theoretically, this schema might not work for bigquery due to the common use of keyfiles instead of user/password credentials, but we can ignore that for now.

KacpiW: Yes, actually here if we'd decide to develop for other systems than Snowflake and Postgres I think we could give it a try and overall drop the idea of schema check using these json files. Right now we have to maintain type/config integrity in two places, here and in Configuration class (and other related classes), which is a bit unhandy and prone to make mistakes. We might want to investigate and try to refactor this into a single class and use PyDantic model validation package that would both read the data from config file and validate the config schema and types. I was thinking to do it here, but I think it needs a bit more time to properly plan it beforehand.

snowshu/adapters/target_adapters/base_remote_target_adapter.py
KacpiW: Well, current state of getting credentials is happening in ConfigurationParser().from_file_or_path() and unfortunately it's different for source and target adapters. Source is doing it while running _build_adapter_profile() (assumes it's always remote source) based on passed credpath, however target is using a custom method (currently the one defined here and the other defined in BaseLocalTargetAdapter and always assumes local target), where the local is having that assumption that creds are always "snowshu". I think we could scope out a ticket to refactor it and possibly contain just one method for creating adapters (not two as we have now) and then wrap somehow that if target credentials are now passed we set "snowshu" as default ones, or something similar for simplicity. Unless I go to far with it and you see a simpler solution?

snowshu/core/configuration_parser.py
KacpiW: That's correct, I guess it's correlated with the other question regarding import of ConfigurationParser(), so if we solve the other one we could revert this change.

snowshu/adapters/target_adapters/base_remote_target_adapter.py
eculicny: Yeah, that makes sense to me since locally we don't worry about non-default credentials.

I'd leave it mostly as is for now, but let's separate the work out for sometime later to rework the config and credentials handling. My only quick suggestion would be moving the global var for the schema json file to the config parser.

KatzperWoz: https://health-union.atlassian.net/browse/DATA-7221

glysav (approved):

snowshu/templates/credentials_schema_target.json
```diff
@@ -0,0 +1,52 @@
...
+                "type": "string"
+              },
+              "user": {
+                "type": "string"
+              }
+            },
+            "required": [
+              "account",
+              "adapter",
+              "database",
```
eculicny, line 35: We might not actually need `database` here since snowshu fully qualifies everything. We can always remove it later if unneeded.

eculicny (approved):

PR #168, opened 2024-04-29
Refactor/refactor replica factory

KatzperWoz:


glysav (approved):

snowshu/adapters/target_adapters/base_local_target_adapter.py
```diff
@@ -89,15 +91,44 @@ def _init_image(self, source_adapter_name: str) -> None:
     def _target_database_is_ready(self) -> bool:
         return self.container.exec_run(self.DOCKER_READY_COMMAND).exit_code == 0
 
-    def finalize_replica(self) -> None:
+    def finalize_replica(self, config: Configuration, **kwargs) -> None:
         """Converts all containers to respective replicas,
         creates 'latest' if any on the running containers are of local arch
         """
-        logger.info("Finalizing target container into replica...")
+        # Section 1: Enable cross-database links if supported
```
eculicny, line 98: Theoretically, the cross-DB and extension stuff might not be needed in all potentially supported cases, but I think the trade-off for simplicity of one finalize step is fine here since we only have a PG local adapter.

eculicny (approved):

PR #169, opened 2024-06-06
Feat/data 7090 snowshu default role

KatzperWoz:
https://health-union.atlassian.net/jira/software/c/projects/DATA/boards/17?assignee=6372bb033c26ca7fa0d09055&selectedIssue=DATA-7090

### DIFF
--- a/snowshu/adapters/snowflake_common.py
+++ b/snowshu/adapters/snowflake_common.py
@@ -8,8 +8,16 @@ from sqlalchemy.pool import NullPool
 
 logger = logging.getLogger(__name__)
 
+SNOWFLAKE_ACCOUNT_DEFAULT_ROLE = "SNOWSHU_REPLICA_BUILDER_ROLE"
 
 class SnowflakeCommon():
+
+    def set_default_role(self) -> None:
+        """Sets the default role for the connection."""
+        if self.credentials.role is None:
+            logger.info('Setting default role to %s...', SNOWFLAKE_ACCOUNT_DEFAULT_ROLE)
+            self.credentials.role = SNOWFLAKE_ACCOUNT_DEFAULT_ROLE
+            logger.debug("Default role set to %s.", self.credentials.role)
 
     def get_connection(
         self,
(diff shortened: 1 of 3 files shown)


snowshu/adapters/snowflake_common.py
```diff
@@ -54,4 +63,6 @@ def _build_conn_string(self) -> str:
         ]
         get_string = "?" + "&".join(get_args) if get_args else ""
 
+        logger.info(f"#### Connection String #### {base_conn}{schema}{get_string}")
```
glysav, line 66: I believe this was added for debugging and is not necessary anymore. Have to be removed since is outputs sensitive info(credentials)

glysav (approved):
Looks good to me. 
The only one concern that I have -- a failure in case of SRB role is not granted to the user. 

Would it make any sense to add a _check_ whether user has Snowshu Replica Builder role? In case not  -- we could use _user's default role_ instead of "_snowshu's default role_". Why I am not sure about such functionality, because it makes snowshu usage easier(more self-determined), but in the same time it adds some _mechanics_ which will be unknown to the user so might confuse the user in some cases

snowshu/adapters/snowflake_common.py
KatzperWoz: definitely, i completly missed it ;<

KatzperWoz:
> Looks good to me. The only one concern that I have -- a failure in case of SRB role is not granted to the user.
> 
> Would it make any sense to add a _check_ whether user has Snowshu Replica Builder role? In case not -- we could use _user's default role_ instead of "_snowshu's default role_". Why I am not sure about such functionality, because it makes snowshu usage easier(more self-determined), but in the same time it adds some _mechanics_ which will be unknown to the user so might confuse the user in some cases

I checked how it behaves right now and in case of not being granted SRB role to user, you're gonna get quite clear message about it and to be honest I'm more afraid of this self-determined behavior and the fact that we're modifing a database here so the user should be sure that it's using the user-role as it's defined in credentials file, rather than changing it for them which might bring more confusion at the end  as you're gonna be thinking that you're going to use your user-role and in fact you're gonna not 🙃

Gal40n04ek (approved):

PR #170, opened 2024-06-11
Feat/data 7091 snowflake create database method

KatzperWoz:


snowshu/adapters/target_adapters/snowflake_adapter/snowflake_adapter.py
```diff
@@ -45,8 +50,13 @@ def initialize_replica(self, config: Configuration, **kwargs):
...
         pass
 
-    def create_database_if_not_exists(self, database):
-        pass
+    def create_database_if_not_exists(self, database: str, **kwargs):
+        replica_name = self.replica_meta["name"].upper()
+        database_name = f"SNOWSHU_{kwargs['uuid']}_{replica_name}_{database}"
+
+        logger.debug(f"Creating database {database_name}...")
+        self.conn.execute(f"CREATE DATABASE IF NOT EXISTS {database_name}")
```
glysav, line 58: Should we handle case when this command fails, or it is handled upstream? I assume if user set role in config it might not have permissions for such command

glysav (approved):

snowshu/adapters/target_adapters/snowflake_adapter/snowflake_adapter.py
KatzperWoz: good suggestion, i've added error handling

Gal40n04ek (approved):

PR #174, opened 2024-06-24
Feat/data 7095 create table

KatzperWoz:


snowshu/adapters/target_adapters/postgres_adapter/postgres_adapter.py
```diff
@@ -151,6 +156,33 @@ def create_schema_if_not_exists(
             else:
                 raise sql_errs
 
+    def create_insertion_arguments(self, relation: Relation, data: Optional[DataFrame] = None) -> dict:
```
ArtemHU, line 159: the return type should be corrected

snowshu/adapters/target_adapters/snowflake_adapter/snowflake_adapter.py
```diff
@@ -146,8 +149,8 @@ def rollback_database_creation(self, databases: Optional[set] = None):
     def create_or_replace_view(self, relation):
         pass
 
-    def create_schema_if_not_exists(self, database, schema, **kwargs):
-        database_name = self.create_database_name(database, kwargs["uuid"])
+    def create_schema_if_not_exists(self, database, schema):
+        database_name = self.create_database_name(database, self.uuid)
```
ArtemHU, line 153: is it available on the instance level, the uuid?

KatzperWoz: The uuid is created on the `GraphSetRunner` object and later distributed across all `target_adapter` objects we create while running process on every graph/thread. I found this as the easiest solution to distribute the same value across all different objects instead of **kwargs that implements more complexity as we need to remember to modify also other target adapters (e.g. postgres) to accept **kwargs as well, which are not utilized afterwards.

KatzperWoz: fixed

ArtemHU (approved):

glysav (approved):

PR #180, opened 2024-09-02
DATA-7099: add table cloning, adjust tests

KatzperWoz:
During development, I encountered several obstacles while following the instructions in the ticket. Implementing the solution as described in the ticket would have required extensive modifications to various Snowshu components, including the compiler. This would have resulted in significantly more work and refactoring than originally estimated.

To avoid this additional effort, I decided to take a different approach. Instead, I applied extra functionality on top of the existing .create_and_load_relation method. This allowed me to achieve the desired outcome more efficiently.

### DIFF
--- a/snowshu/adapters/target_adapters/base_target_adapter.py
+++ b/snowshu/adapters/target_adapters/base_target_adapter.py
@@ -97,8 +97,8 @@ class BaseTargetAdapter(BaseSQLAdapter):
         """
         if clone:
             self.clone_relation(relation)
-            return 
-        
+            return
+
         if data is None and relation.data.empty:
             logger.warning(
                 "Both data and relation.data are empty for %s. "
@@ -132,7 +132,7 @@ class BaseTargetAdapter(BaseSQLAdapter):
             raise
 
         logger.info(final_message)
-        
+ 
     def clone_relation(self, relation: Relation) -> None:
         """Clones a relation in the target.
         Args:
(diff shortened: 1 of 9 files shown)


snowshu/adapters/target_adapters/base_target_adapter.py
```diff
@@ -129,6 +133,18 @@ def load_data_into_relation(
...
         logger.info(final_message)
 
+    def clone_relation(self, relation: Relation) -> None:
+        """Clones a relation in the target.
+        Args:
+            relation: The relation to clone.
+        """
+        database_name = self.create_database_name(relation.database)
+        clone_query = f"""
+            CREATE TABLE IF NOT EXISTS {database_name}.{relation.schema}.{relation.name} AS
```
glysav: Shouldn't this be create or replace?

KatzperWoz: I'm gonna change it for now, as create or replace gonna give us always the latest subset, tho cannot think of scenario where it is going to have an advantage.  Let me know if you have any 🤔

glysav: I was only wondering whether it is possible this query to fail in case of table already exists(not sure if such scenario possible, but still...)

KatzperWoz: I think in such scenario the query is ignored and process continues

glysav (approved):

ArtemHU (approved):

PR #181, opened 2024-09-09
DATA-7101: promote to prod cli functionality

KatzperWoz:


snowshu/adapters/target_adapters/snowflake_adapter/utils.py
```diff
@@ -0,0 +1,76 @@
...
+import snowflake.connector
+
+
+logger = logging.getLogger(__name__)
+
+
+def connect_to_database(credentials):
+    try:
+        conn = snowflake.connector.connect(
+            user=credentials["user"],
```
ArtemHU: potentially we can use the unpacking technic here **

snowshu/adapters/target_adapters/snowflake_adapter/utils.py
```diff
@@ -0,0 +1,76 @@
...
+            password=credentials["password"],
+            account=credentials["account"],
+            database=credentials["database"],
+            role=credentials["role"],
+        )
+        return conn
+    except snowflake.connector.errors.Error as e:
+        logger.info(f"Failed to connect to Snowflake: {e}")
```
ArtemHU: should it be logger.error?

snowshu/adapters/target_adapters/snowflake_adapter/utils.py
```diff
@@ -0,0 +1,76 @@
...
+        return None
+
+
+def rename_database(cursor: snowflake.connector.Cursor, old_name: str, new_name: str):
+    """Rename a database, dropping the new name if it already exists."""
+    try:
+        cursor.execute(f"ALTER DATABASE {old_name} RENAME TO {new_name}")
```
ArtemHU, line 25: can we try to execute both statements in one transaction?
drop if exists + actual rename

snowshu/adapters/target_adapters/snowflake_adapter/utils.py
```diff
@@ -0,0 +1,76 @@
...
+            raise e
+
+
+def handle_exisiting_prod_databases(
+    cursor: snowflake.connector.Cursor,
+    prod_prefix: str,
+    current_date: pendulum.datetime = pendulum.now(),
+):
+    """Rename existing production databases."""
+    cursor.execute("SHOW DATABASES LIKE '{prod_prefix}%'")
```
ArtemHU: the like condition can be replaced with` STARTS WITH`

snowshu/core/main.py
```diff
@@ -182,3 +186,86 @@ def launch_docker_cmd(replica: str):
...
+@click.option(
+    "--credentials-file",
+    default="./replicas/credentials.yml",
+    type=click.Path(exists=True),
+    help="The path to the credentials file.",
+    required=True,
+)
+@click.option(
+    "--prod-prefix",
+    default="SNOWSHU_PROD",
```
ArtemHU, line 216: Should we specify possible values in advance to prevent scenarios where an incorrect database is provided?

snowshu/adapters/target_adapters/snowflake_adapter/utils.py
KatzperWoz: potentially we could, but I think current approach is more explicit and easier to understand and debug in case of anything, if I give both statements I won't get any return if we actually dropped something cause - we could dig into cursor return object, but it will take time now and I don't really see any particular profit from it, what do you think? maybe I'm missing something @ArtemHU

snowshu/core/main.py
KatzperWoz: didn't think about it, will add choice

snowshu/adapters/target_adapters/snowflake_adapter/utils.py
ArtemHU: the end result will remain consistent in both scenarios, even though the log provides sufficient information to comprehend what was dropped.

ArtemHU (approved):

glysav (approved):

PR #182, opened 2024-09-23
Feat/developer workflow enhancement

KatzperWoz:
1. Replaced ```prospector``` with current market standard [ruff](https://docs.astral.sh/ruff/). It's also like 1000x times faster than prospector, which improves ci/cd.
2. Added file formatting as pre-commit hook and ruff linter as pre-push hook. It alleviate developer work, comparing to prospector
3. Formatted files with ruff. No logical changes has been made.


setup.py
```diff
@@ -2,33 +2,40 @@
 
 from setuptools import find_packages, setup
 
-VERSION = '0.0.3'
-PYTHON_REQUIRES = '3.8'
+VERSION = "0.0.3"
+PYTHON_REQUIRES = "3.8"
```
ArtemHU, line 6: why 3.8, should we bump it up?

setup.py
```diff
@@ -2,33 +2,40 @@
...
-    packagedata['long_description'] = readme.read()
-    packagedata['long_description_content_type'] = 'text/markdown'
+with open("./README.md") as readme:
+    packagedata["long_description"] = readme.read()
+    packagedata["long_description_content_type"] = "text/markdown"
 
-for file_name in ['base.txt', 'snowflake_pins.txt']:
-    with open(f'./requirements/{file_name}') as requirements:
+for file_name in ["base.txt", "snowflake_pins.txt"]:
+    with open(f"./requirements/{file_name}") as requirements:
```
ArtemHU, line 36: it's recommended to provide encoding rather than rely on default one

ArtemHU (approved):

mike-ob (approved):

setup.py
KatzperWoz: I'll create separate ticket for it, i was focusing solely on development workflow here

